PR #117, opened 2015-01-07
Make core packages null by default instead of excluding them

shlevy:
This allows overriding core packages

### DIFF
--- a/src/Cabal2Nix/Normalize.hs
+++ b/src/Cabal2Nix/Normalize.hs
@@ -4,7 +4,6 @@ module Cabal2Nix.Normalize ( normalize, normalizeList ) where
 
 import Distribution.NixOS.Derivation.Cabal
 import Cabal2Nix.Name
-import Cabal2Nix.CorePackages
 import Data.List
 import Data.Char
 import Data.Function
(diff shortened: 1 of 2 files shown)


shlevy:
Note that before updating `nixpkgs` to use this you must set `pretty = null` in `haskell-packages.nix` otherwise the agda `pretty` package is used instead.

shlevy:
@peti We have a somewhat pressing need to use this, I'd appreciate a quick review if possible.

src/Cabal2Nix/CorePackages.hs
```diff
@@ -8,7 +10,7 @@ module Cabal2Nix.CorePackages ( corePackages, coreBuildTools ) where
 -- The commented packages have previously been non-core, so we don't filter
 -- them.
 corePackages :: [String]
-corePackages = [
+corePackages = map toNixName [
```
peti: What is the purpose of this change? Currently, `corePackages` contain a list of Haskell package names, *not* Nix attributes. Since the latter can be generated out of the former -- but not vice versa -- it seems like a bad idea to apply `toNixName` here.

src/Cabal2Nix/CorePackages.hs
```diff
@@ -37,7 +39,7 @@ corePackages = [
   ]
 
 coreBuildTools :: [String]
-coreBuildTools = [
+coreBuildTools = concatMap buildToolNixName [
```
peti: Ditto.

src/Cabal2Nix/CorePackages.hs
shlevy: The lists are only used in one place, both before and after this patch. But sure, I'll move this closer to where it's used (when I'm back at my computer)

peti:
Are you aware of the fact that `corePackages` will be empty altogether in v2.0, https://github.com/NixOS/cabal2nix/commit/8a5ec344e72250b8ccdbb34a311148fcd0c23135?

peti:
I'd rather remove the filtering of core packages altogether, like it's already done in the v2.x branch, and provide `null` overrides for all of them rather than relying on default null values in every expression.

src/Cabal2Nix/CorePackages.hs
shlevy: Fixed this in force push

shlevy:
No, I didn't know that. What's the time frame for switching to 2.0?

peti:
Some time later today.

shlevy:
Ah, cool! Perfect then, I'll close this.

Does `2.0` work on non-haskell-ng?

peti:
The derivation format in NG is different; so 2.x cannot be used to maintain the old stuff.

PR #127, opened 2015-01-13
Add special case for optparse-applicative 0.10.

teh:
I tried updating elm instead but it's not entirely trivial so this is
a quick fix until elm has been updated by its maintainers.


peti:
Please include a comment with the entry that documents because of which package it was added. This makes it easier to decide later whether those entries can be removed or not.

teh:
Does this work?

PR #131, opened 2015-01-15
Map system library name "libsystemd" to "systemd" for Nixpkgs

ocharles:
On a somewhat related note - would anyone mind if I had commit rights to `cabal2nix`?


shlevy:
@peti can grant you access I think.

peti:
Done. You should now be able to commit.

PR #141, opened 2015-01-26
bytestring-arbitrary: Add "llvm" extraLib

richardlarocque:
Warning: I have not yet built or run this code.  Apologies in advance if it's junk.

bytestring-arbitrary from nixpkgs' haskell NG doesn't build for me.  The error makes me think it has something to do with the -fllvm option in its [cabal file](https://github.com/tsuraan/bytestring-arbitrary/blob/master/bytestring-arbitrary.cabal).

Addings pkgs.llvm to the extraLibs in its hackage-packages.nix definition fixed the issue for me.  But that's a generated file.  I can't commit those changes upstream directly.  I'm trying to commit the equivalent changes here instead.

### DIFF
--- a/src/Cabal2Nix/PostProcess.hs
+++ b/src/Cabal2Nix/PostProcess.hs
@@ -26,6 +26,7 @@ postProcess' deriv@(MkDerivation {..})
   | pname == "bits-extras"      = deriv { configureFlags = Set.insert "--ghc-option=-lgcc_s" configureFlags
                                         , extraLibs = Set.filter (/= "gcc_s") extraLibs
                                         }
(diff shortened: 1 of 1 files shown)


peti:
You are right; `bytestring-arbitrary` was missing `llvm` to build. That issue was fixed a few days ago in https://github.com/NixOS/nixpkgs/commit/cdb43faa3da84fb41d22d2cb45698d9f7cbf80e2#diff-8f392bd0e4fd42a66d9e0cd4901ec287R364. Just update to a recent version of Nixpkgs, then it shoud build fine.

PR #143, opened 2015-02-05
update post-processing for liquid-fixpoint and liquidhaskell

gridaphobe:
liquid-fixpoint needs the `-fbuild-external` flag to force building the
ocaml binary locally instead of using a precompiled binary (not
guaranteed to work in nixos AFAIK)

### DIFF
--- a/src/Cabal2Nix/PostProcess.hs
+++ b/src/Cabal2Nix/PostProcess.hs
@@ -72,7 +72,8 @@ postProcess' deriv@(MkDerivation {..})
   | pname == "leksah-server"    = deriv { buildDepends = Set.insert "process-leksah" buildDepends }
   | pname == "lhs2tex"          = deriv { extraLibs = Set.insert "texLive" extraLibs, phaseOverrides = lhs2texPostInstall }
   | pname == "libffi"           = deriv { extraLibs = Set.delete "ffi" extraLibs }
-  | pname == "liquid-fixpoint"  = deriv { buildTools = Set.insert "ocaml" buildTools }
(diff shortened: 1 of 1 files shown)


src/Cabal2Nix/PostProcess.hs
```diff
@@ -72,7 +72,8 @@ postProcess' deriv@(MkDerivation {..})
   | pname == "leksah-server"    = deriv { buildDepends = Set.insert "process-leksah" buildDepends }
   | pname == "lhs2tex"          = deriv { extraLibs = Set.insert "texLive" extraLibs, phaseOverrides = lhs2texPostInstall }
   | pname == "libffi"           = deriv { extraLibs = Set.delete "ffi" extraLibs }
-  | pname == "liquid-fixpoint"  = deriv { buildTools = Set.insert "ocaml" buildTools }
+  | pname == "liquid-fixpoint"  = deriv { buildTools = Set.insert "ocaml" buildTools, configureFlags = Set.insert "-fbuild-external" configureFlags }
+  | pname == "liquidhaskell"    = deriv { doCheck = False } -- test-suite requires cvc4 or z3
```
peti: Maybe we should add `cv4` as a build input to make the tests succeed?

gridaphobe: That's a pretty substantial buildInput and I'd rather let users choose which SMT solver they want to use. Is it possible to add a buildInput that will only be realized if `doCheck` is enabled?

peti:
All inputs listed in `testDepends` will be pulled in only when `doCheck` is true.

gridaphobe:
Ah right, and I guess it's fine to add non-haskell packages to `testDepends` in the nix expression since we're not generating a cabal file from it.

peti:
Yes, exactly.

Issue #155, opened 2015-03-30
List dependencies for generate-cabal-file.hs


abbradar:
It's not trivial to start hacking/building `cabal2nix` because you can't generate Cabal file without some dependencies. It would be nice to store a simple .cabal somewhere, like that:
```
name:                generate-cabal-file
version:             0.1.0.0
build-type:          Simple
license:             BSD3
license-file:        LICENSE
build-type:          Simple
cabal-version:       >=1.10

executable generate-cabal-file
  main-is:             generate-cabal-file.hs
  build-depends:       base
                     , cartel
                     , mtl
                     , process
  default-language:    Haskell2010
```
just to ease the process -- with it one can just `cabal2nix --shell` for bootstrapping. A .nix derivation would work too, I guess.

peti:
Running

    nix-shell --pure release.nix -A cabal2nix.ghc784.x86_64-linux.env --command "runhaskell generate-cabal-file.hs >cabal2nix.cabal"


should do the trick.

abbradar:
I haven't thought of using `release.nix` for this, thanks! Maybe add this to README?

peti:
Yes, the build process should definitely be documented. I was too lazy so far, I'll tackle this issue soon.

PR #156, opened 2015-03-30
add bootstrap.nix to ease .cabal file generation

abbradar:
I went ahead and made a possible solution for #155. This is basically auto-generated with `cabal2nix --shell` with a changed `shellHook`.


abbradar:
Should be fixed by describing process in README instead.

PR #168, opened 2015-05-22
add `z3` to the buildTools for liquid-fixpoint and liquidhaskell so we can run the tests

gridaphobe:
NB: `z3` takes a while to build (and liquidhaskell has a large testsuite too). Should we mention this somewhere so people can be aware and possibly choose to disable tests locally?


peti:
Hmm, not sure where we could mention that so that people actually see the hint?

PR #174, opened 2015-06-18
Write error messages for parse failures to stderr

supki:
Currently, when running `cabal2nix . > foo.nix` against a malformed cabal file, the error message ends up in `foo.nix` and `cabal2nix` appears to fail silently, which is not ideal.


peti:
Sorry for the delay!

PR #176, opened 2015-06-25
Remove xmonad-screenshot from broken packages

dmalikov:
This package is seem to be stable now

### DIFF
--- a/src/hackage2nix.hs
+++ b/src/hackage2nix.hs
@@ -3921,7 +3921,6 @@ defaultConfiguration = Configuration
     , "xmonad-bluetilebranch"
     , "xmonad-contrib-bluetilebranch"
     , "xmonad-eval"
-    , "xmonad-screenshot"
     , "xmpipe"
     , "XMPP"
     , "xournal-builder"


shlevy:
Why does cabal2nix even need to know which packages are broken? Can't this list be maintained in nixpkgs directly?

peti:
@dmalikov, I cannot compile `xmonad-screenshot`. The build fails as follows:

~~~
Preprocessing library xmonad-screenshot-0.1.1.0...
[1 of 2] Compiling XMonad.Util.WorkspaceScreenshot.Internal ( src/XMonad/Util/WorkspaceScreenshot/Internal.hs, dist/build/XMonad/Util/WorkspaceScreenshot/Internal.o )

src/XMonad/Util/WorkspaceScreenshot/Internal.hs:41:3:
    Non type-variable argument in the constraint: MonadState XState m
    (Use FlexibleContexts to permit this)
    When checking that ‘workspaceIdToWindowSpace’ has the inferred type
      workspaceIdToWindowSpace :: forall (m :: * -> *).
                                  MonadState XState m =>
                                  WorkspaceId -> m (S.Workspace WorkspaceId (Layout Window) Window)
    In an equation for ‘captureWorkspacesWhen’:
        captureWorkspacesWhen p hook
          = captureWorkspacesWhenId (workspaceIdToWindowSpace >=> p) hook
          where
              workspaceIdToWindowSpace i
                = gets
                  $ head . filter (\ w -> S.tag w == i) . S.workspaces . windowset
~~~~

Does the build work for you?

dmalikov:
> Preprocessing library xmonad-screenshot-0.1.1.0...

@peti could you please try `0.1.2` version from package?

PR #177, opened 2015-06-27
Add z3 to the DYLD_LIBRARY_PATH on darwin

gridaphobe:
It fails to build with

```
Executing: dist/build/Z3/Base/C_hsc_make  >dist/build/Z3/Base/C.hs
dyld: Library not loaded: libz3.dylib
  Referenced from: dist/build/Z3/Base/C_hsc_make
  Reason: image not found
running dist/build/Z3/Base/C_hsc_make failed (exit code -5)
command was: dist/build/Z3/Base/C_hsc_make  >dist/build/Z3/Base/C.hs
```

otherwise.


peti:
Uh, did you actually test this patch?

gridaphobe:
I tested it by editing hackage-packages.nix in nixpkgs.

Did I do something stupid like a type error on the Haskell end?

peti:
The Derivation type in cabal2nix has no field "preBuild". All those pre/post fields are lumped together in "phaseOverrides", so the patch would not compile. I suppose I should really re-enable Travis-Ci for this repository to catch that kinds of stuff. A had it running a while ago, but then the build broke when I started requiring ghc-7.10.1, which Travis didn't have at the time. Duh.

gridaphobe:
Ack, so sorry about that!

I agree, Travis-CI would be very nice, especially for one-line edits
like this that are easily made via GtiHub's web-editor :)

PR #178, opened 2015-06-28
Fix build for haskell package 'gf'

k0001:
The building of the `gf` Haskell package is a bit special. Not only it builds the `gf` compiler and related Haskell libraries, but also a standard library for the GF (Grammatical Framework) programming language, which are themselves compiled with the `gf` compiler. 

This change allows for the full `gf` package to be built.

### DIFF
--- a/src/Cabal2Nix/PostProcess.hs
+++ b/src/Cabal2Nix/PostProcess.hs
@@ -51,6 +51,8 @@ postProcess' deriv@(MkDerivation {..})
   | pname == "gtk2hs-buildtools"= deriv { buildDepends = Set.insert "hashtables" buildDepends }
   | pname == "haddock" && version < Version [2,14] []
                                 = deriv { buildTools = Set.insert "alex" (Set.insert "happy" buildTools) }
+  | pname == "gf"               = deriv { phaseOverrides = gfPhaseOverrides
+                                        , doCheck = False }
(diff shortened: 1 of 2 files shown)


src/Cabal2Nix/PostProcess.hs
```diff
@@ -51,6 +51,8 @@ postProcess' deriv@(MkDerivation {..})
   | pname == "gtk2hs-buildtools"= deriv { buildDepends = Set.insert "hashtables" buildDepends }
   | pname == "haddock" && version < Version [2,14] []
                                 = deriv { buildTools = Set.insert "alex" (Set.insert "happy" buildTools) }
+  | pname == "gf"               = deriv { phaseOverrides = gfPhaseOverrides
+                                        , doCheck = False }
```
k0001: Some tests fails in the upstream distribution, so I disabled them here.

k0001:
Also, `hackage-packages.nix` in `nixpkgs` lists version `3.6` for `gf`, but it should be upgraded to `3.7`. Do I need to do something special to get that or will version `3.7` be there next time `hackage-packages.nix` is generated?

peti:
The latest version is already in the `haskell-updates` branch; it will be merged to `master` automatically soon; you don't have to worry about that.

PR #188, opened 2015-08-04
Try keeping an older version of tar around.

mdorman:
7.0-7.4 don't have bytestring recent enough to build with the new version of tar, so hydra.cryp.to is showing failures for some packages with those versions. I think this is the necessary incantation to keep the most recent version < 0.4.2 around---at least it mirrors what aeson does.  Then with strategic overrides, I believe, we can get the older versions building again.


peti:
I've made that change in 012eb638a5db2ba7cf2c94e4375ea5cc05aefde0 already.

PR #190, opened 2015-08-15
Fix package lookup for shadowed system packages

bennofs:
System packages which have the same package name as haskell packages should be brought
into scope by using `inherit systemPackageName;`. Previously, `hackage2nix` would
generate `systemPackageName = null` or nothing at all instead.

The change introduces a slight problem for executables that depend on themselves as a
tool (c2hs' testsuite for example depends on the c2hs tool). To fix this, such a
dependency is now removed in the normalization step of cabal2nix.


bennofs:
I also verified that this generates the exact same package set as the previous cabal2nix, except in the mentioned case of `c2hs`.

peti:
Thank you very much!

Issue #184, opened 2015-07-19
Expose hackage2nix as a library


Fuuzetsu:
Use case: overriding the default configuration to be able to generate a more fine-grained package sets with the cabal-provided versions on per-project basis.

teh:
@Fuuzetsu would this allow to e.g. follow the stackage set of packages? Or to have a specific one for yesod?

Fuuzetsu:
@teh with further work, yes, that's my motivation for this request

nrolland:
that would be really good. having the benefits of nix immutability and the curated set of stockage would be really awesome.

peti:
I agree that `hackage2nix` should be re-factored so that it's easier to re-use in external code. I believe that's a good idea mostly because it would help us clean up the code, which is fairly monolithic and messy at the moment.

I don't believe, however, that having `hackage2nix` available as a library is important for implementing support for Stackage. In fact, it's quite simple to track Stackage already: Download https://www.stackage.org/lts/cabal.config, edit the file slightly for syntax (or have a `sed` script do it automatically), add its contents to the [`defaultPackageOverrides`](https://github.com/NixOS/cabal2nix/blob/badda1eb05994906083085686dbdd86f13237e41/src/hackage2nix.hs#L301) list, and re-run `hackage2nix`. That would give you a Haskell package set based on Stackage.

Now, that approach isn't quite good enough for our purposes in Nixpkgs, because we would like to have Stackage support *in addition* to the bleeding-edge package set we currently provide. This means that the Stackage version list must go into [`extraPackages`](https://github.com/NixOS/cabal2nix/blob/badda1eb05994906083085686dbdd86f13237e41/src/hackage2nix.hs#L336) instead, to make sure `hackage2nix` provides all the package versions we need for Stackage even though some of those are older than the latest available version from Hackage. Finally -- and this is the only part that needs actual coding -- `hackage2nix` must be extended to generate a set of overrides, a `configuration-stackage.nix` file, if you will, that modifies our default choices (latest version) to the Stackage choice where appropriate.

I believe that this whole feature can be built into `hackage2nix` fairly easily; I see no need to have a separate tool for that purpose.

bennofs:
That is one way to implement support for stackage. However, if we go that way, users can't easily generate their own package set, since then they would have to edit the global hackage-packages.nix. So, what about the following:

* stackage2nix is a tool that generates a file `stackage-packages.nix`, just like `hackage-packages.nix`
[...]

PR #191, opened 2015-08-18
hackage2nix as a library

bennofs:
This PR splits the parts of hackage2nix which are responsible for generating the package set in a library, fixing #184. I'm sorry if I have duplicated work (I didn't see #184 was assigned until after I already worked on trying to write a stackage2nix). 

I also made the configuring phase (deciding whether to jailbreak, enable tests, etc) a bit more flexible. For example, stackage already contains information on whether to run the tests or not, so it makes sense to be able to directly use of that.

The more flexible package configuration also allowed me to fix a bug in package generation: some packages where needlessly marked as broken, because there tests depend on unavailable packages. The new behavior is that in such a case, the package is not marked as broken, but the tests are disabled. hackage2nix will now try the following configurations, in this order (the first configuration to succeed is used):

* Tests enabled, no jailbreak
* Tests enabled, jailbreak
* Tests disabled, no jailbreak
* Tests disabled, jailbreak

If all of that fails, a broken derivation will be generated. 

The PR also includes some fixes to make hackage2nix use less memory.

### DIFF
--- a/generate-cabal-file.hs
+++ b/generate-cabal-file.hs
@@ -70,7 +70,7 @@ commonBuildDepends =
     , unconstrained "hackage-db"
     , unconstrained "lens"
     , unconstrained "monad-par"
-    , unconstrained "monad-par-extras"
+    , unconstrained "abstract-par"
     , unconstrained "mtl"
     , unconstrained "optparse-applicative"
     , package "pretty" (gtEq [1,1,2])
--- a/generate-cabal-file.hs
+++ b/generate-cabal-file.hs
@@ -70,7 +70,7 @@ commonBuildDepends =
     , unconstrained "hackage-db"
     , unconstrained "lens"
     , unconstrained "monad-par"
-    , unconstrained "abstract-par"
+    , unconstrained "monad-par-extras"
     , unconstrained "mtl"
     , unconstrained "optparse-applicative"
     , package "pretty" (gtEq [1,1,2])
--- a/src/Cabal2Nix/HackageGit.hs
+++ b/src/Cabal2Nix/HackageGit.hs
@@ -10,8 +10,9 @@ import Data.ByteString.Lazy.Char8 ( fromStrict )
 import Data.Digest.Pure.SHA ( sha256, showDigest )
 import Data.List ( foldl' )
 import Data.Map
-import Data.Maybe ( fromMaybe )
+import Data.Maybe ( fromMaybe, mapMaybe )
 import Data.String.UTF8 ( toString, fromRep )
+import Distribution.Package
 import Distribution.PackageDescription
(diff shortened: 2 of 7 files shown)


peti:
You might be interested in https://github.com/NixOS/cabal2nix/issues/184#issuecomment-132592633, too.

bennofs:
Regarding memory usage, here are two heap profiles (unfortunately, hp2ps does not seem to have a way to make the colors of both images match).
### Master
master
### With the changes in the PR
new

While the second one looks about constant, the first one is leaking. There are some performance bugs in the cabal flag assignment (the whole first yellow triangle is really cabal trying to finalize a single package with many flags), which show up as spikes in the second graph.

bennofs:
I'll try to see if I can find a way without changing the HackageGit representation.

peti:
What happens when you specify an upper limit on the heap? Does that improve the situation?

bennofs:
@peti I can try that

bennofs:
First, here's what happens when I just remove the IO on my branch. The leak is better visible here:

hackage2nix leak

bennofs:
At least with `-M200M`, it fails with a heap exception, but that might be not enough memory...

bennofs:
It even exhausts the heap with `-M700M`

bennofs:
I give up trying to make this use less memory without changing the hackage db representation. It seems impossible to convince ghc to garbage collect the hackage map while the output is produced.

peti:
I suppose this means that I'll have to consider changing the API of `hackage-db`. Having that lazy map seemed like an obvious choice for parsing that tarball from `cabal-install`, but in retrospect that API isn't so good anymore because it's forces significant costs on users who don't want to have the `GenericPackageDescription`, but who really just want to see the set of available packages.

peti:
I've been thinking about your extended algorithm that potentially disables the test suite in an attempt to salvage the build. Is that behavior desirable? I mean, I want the test suite for my packages to run, and if that's not possible for whatever reason, then I want the build to *fail*. This is essential for my use of Hydra/Nix as a Continuous Integration tool. Now, if `hackage2nix` would decide to disable the test suite on one of my packages in an attempt to make a build succeed that would otherwise fail, then I'd be rather annoyed, actually.

How do you feel about that issue?

bennofs:
I agree with you on this. The issue again is that the package itself and the test run of it aren't separate: as an end-user, I might want to choose not to run the tests, but still use the binary cache for the library itself.

Note however that tests are *only* disabled when the test suite depends on library that is not available in nixpkgs. IMO, that means that the tests are broken, not the package as a whole. But that's currently not expressible in nixpkgs, so I'm fine with either way.

peti:
I suppose this PR has become obsolete.

PR #207, opened 2015-09-15
Replace fromMaybe&Map.lookup with Map.findWithDefault

DNNX:
Just a tiny refactoring.


peti:
Duh, there is a function `Map.findWithDefault`?!? Very nice. Thanks for letting me know. :-)

PR #223, opened 2016-04-03
Aeson instance update

mdorman:
I'm not 100% sure that these are the right instances definitions, but perhaps they could at least be a starting point?

### DIFF
--- a/distribution-nixpkgs/src/Distribution/Nixpkgs/Haskell/FromCabal/Configuration.hs
+++ b/distribution-nixpkgs/src/Distribution/Nixpkgs/Haskell/FromCabal/Configuration.hs
@@ -1,4 +1,6 @@
 {-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE CPP               #-}
+{-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# OPTIONS_GHC -fno-warn-orphans #-}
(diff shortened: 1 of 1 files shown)


peti:
I tried running `hackage2nix` with that patch applied, but then it can no longer parse its config file: 

    hackage2nix: user error (invalid config file at "/home/simons/.nix-defexpr/pkgs/development/haskell-modules/configuration-hackage2nix.yaml")

Another problem is that I'm reluctant to depend on aeson 0.11.x, because that version cannot be found in any LTS Haskell branch yet (and won't be until LTS 6 comes out).

I'm not sure how to proceed.

mdorman:
Yeah, I was suspicious that it wasn't the right thing.  I'll try monkeying around a little more.

Don't we have aeson_0.9 around still, though?  It seems like just forcing that is the right thing for the moment.

mdorman:
So, I just said to heck with trying to simplify them---I just pulled out the common code on the existing instance and otherwise kept them the same---and I conditionalized the whole thing on the aeson version, just for good measure.

mdorman:
I don't know why the travis test failed.  It works for me in a pure shell:

```
mdorman@aching:~/src/cabal2nix/distribution-nixpkgs (master)$ nix-shell --pure -I nixpkgs=/home/mdorman/src/nixpkgs --no-build-output --run "runghc Setup.lhs configure --enable-tests; runghc Setup.lhs build; runghc Setup.lhs test --verbose"
Configuring distribution-nixpkgs-1...
Building distribution-nixpkgs-1...
Preprocessing library distribution-nixpkgs-1...
[ 8 of 20] Compiling Distribution.Nixpkgs.Haskell.FromCabal.Configuration ( src/Distribution/Nixpkgs/Haskell/FromCabal/Configuration.hs, dist/build/Distribution/Nixpkgs/Haskell/FromCabal/Configuration.o ) [dist/build/autogen/cabal_macros.h changed]
In-place registering distribution-nixpkgs-1...
Preprocessing test suite 'spec' for distribution-nixpkgs-1...
Linking dist/build/spec/spec ...
Running 1 test suites...
Test suite spec: RUNNING...
dist/build/spec/spec
Test suite spec: PASS
Test suite logged to: dist/test/distribution-nixpkgs-1-spec.log
1 of 1 test suites (1 of 1 test cases) passed.
```

peti:
The travis build is run via `stack`. That makes a difference, apparently? Not sure what is going on either. I'll give this a try locally.

peti:
See https://github.com/sol/doctest/issues/127#issuecomment-205834514 for details about the issue withh `stack`.

peti:
Merged in baed1b5906b26e999948c6be54296994e2a84689.

PR #227, opened 2016-05-28
hlibgit2: all newer versions should get pkg.git

mdorman:
This package has been broken since haskellPackages.git was introduced a
couple of weeks ago *and* the newer version became the default, because
only this specific rev was getting the pkg.git dependency.  Simply
loosen the dependency to fix.


mdorman:
@peti I'm pretty sure this is the right thing to do, but please let me know if it's not.

PR #229, opened 2016-05-28
Allow building with ghc-8.0.1

mdorman:
As of deepseq-1.4.0, the ability to derive generics has been folded into the main library.  Furthermore, deepseq-1.4.2.0 is included with ghc-8.0.1 (and 1.4.1.1 was bundled with ghc-7.10.1).

Before this change, attempts to compile with ghc-8.0.1 would throw errors like:
```
  src/Distribution/Nixpkgs/Haskell/OrphanInstances.hs:32:10: error:
      Not in scope: type constructor or class ‘NFData’
```
Changing to just use the automatic deriviation seemed like the easiest way to resolve the issue.  I *believe* this will just work on 7.10.x and with the exception of  a couple of terms, might simply work all the way back to the "beginning of time"...or at least when NFData first appeared.

### DIFF
--- a/hackage2nix/hackage2nix.cabal
+++ b/hackage2nix/hackage2nix.cabal
@@ -30,7 +30,7 @@ executable hackage2nix
     , bytestring
     , Cabal >= 1.22.2
     , containers
-    , deepseq-generics
+    , deepseq >= 1.4
     , directory
     , distribution-nixpkgs
     , filepath
--- a/distribution-nixpkgs/distribution-nixpkgs.cabal
+++ b/distribution-nixpkgs/distribution-nixpkgs.cabal
@@ -29,7 +29,7 @@ library
     , bytestring
     , Cabal >= 1.22.2
     , containers
-    , deepseq-generics
+    , deepseq >= 1.4
     , directory
     , filepath
     , hackage-db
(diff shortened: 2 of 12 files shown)


mdorman:
I have confirmed that this still builds with 7.10.3.

mdorman:
@peti When/if this goes in, my changes in the `configuration-common.nix` file to tie this to an older version of deepseq-generic will also need to be removed, since that's just a stopgap.

hackage2nix/hackage2nix.cabal
```diff
@@ -30,7 +30,7 @@ executable hackage2nix
     , bytestring
     , Cabal >= 1.22.2
     , containers
-    , deepseq-generics
+    , deepseq
```
peti: We should probably depend on `deepseq >= 1.4`. Also, note that these Cabal files are generated from `package.yaml` by `stack`/`hpack`, so changes here will be lost.

mdorman:
Doh!  An excellent point about the versioned dependencies.

As for the `stack` stuff, I made the obvious edits, but you've made the NixOS Haskell environment good enough that I've been able to remain utterly `stack` ignorant. :)

peti:
Thank you for the patches!!

PR #230, opened 2016-06-03
Add flag (--impl) for specifying compiler id.

dajmaki:
This is useful for specifying the branch in "if impl(...)" expressions.
Useful for generating nix expression for projects that support both GHC
and GHCJS, but have different dependencies depending on the compiler.


peti:
I  very much like the patch. Can we call the flag `--compiler-id` instead of `--impl`, though? That feels more descriptive.

PR #236, opened 2016-07-18
Fix build environment for gi-* packages

k0ral:
I tried to port the manual changes I did in my local `configuration-common.nix` to have various `gi-*` packages work. There may be mistakes/typos as I haven't had the opportunity to build it, let alone run it.


peti:
This patch doesn't compile according to travis-ci.

k0ral:
It now compiles fine :) .

PR #242, opened 2016-09-06
Another attempt to fix the gi-gst packages

mdorman:


mdorman:
This is almost certainly not the appropriate fix!  I think parts of it are closer, but I can't seem to find a way to actually look at the overrides we expect to see in the full hackage-packages.nix file without running hackage2nix.

mdorman:
So perhaps this is right, or at least workable.

The first observation is that the pkg-config info for the gstreamer-1.0 stuff should map to different names than the 0.10 stuff.  I mapped them to the obvious *-1 names, which, to be clear, *do not exist* in nixpkgs, but being distinct this way seemed most straightforward and umambiguous.

Then I changed the postProcessing code to subtitute a reference for the real path in for the fake path that gets put in above.  I chose this route because it appeared to work in my testing.

That said, my testing may be absolute crap: to test, rather than figure out how to setup hackage2nix, etc., I *instead* changed cabal2nix --shell to include the overrides from the derivation.  I have that as a separate patch, if you'd be interested.

Anyway, this entire change my be utterly wrong; I freely concede that.  I'm just not finding it entirely obvious how to make it better.  I am perfectly happy to try and do the work given some guidance on what you think the appropriate shape of the solution is---or if you just want to do it right yourself, I'm find with that too.

mdorman:
@peti To be clear, please don't merge this until you've looked it over; it's probably no more correct than my last one.

peti:
I'll take a look at those builds ASAP. right now, it's not obvious to me either what the best solution would be. Give me a second ...

mdorman:
Like I said in the email I sent yesterday (did you get that?), I know what the necessary outcome is, it's mostly about figuring out what I need to give to cabal2nix to get the appropriate output.

peti:
Yes, I saw your e-mail. I didn't have a chance to respond right away and then it fell through the cracks. I am sorry.

I believe eb712e908527bf581f54ba72e289fc6a554cc277 fixes the issue. Please let me know if you discovery any other remaining problems.

PR #247, opened 2016-10-24
Make sure that pango gets the appropriate pango dependency

mdorman:
If you do `git blame` on `pkgs/development/haskell-modules/hackage-packages.nix` in `nixpkgs/master`, you'll see this one modification that sticks out like a sore thumb:

```
bccd75094 (Kirill Boltaev        2016-09-12 00:24:51 +0300 124889)      }) {inherit (pkgs.gnome2) pango;};
```

If you look at that line (the end of the `pango` expression) in `peti/haskell-updates`, it has:

```
   }) {inherit (pkgs.gnome) pango;};
```

This fails to evaluate:

```
error: attribute ‘gnome’ missing, at /home/mdorman/src/nixpkgs/pkgs/development/haskell-modules/hackage-packages.nix:126236:19
(use ‘--show-trace’ to show detailed location information)
nix-env failed with error code 1.
```

So it seems clear to me that we need to have `cabal2nix` create the appropriate dependency automatically.  I _think_ this change does that.

### DIFF
--- a/src/Distribution/Nixpkgs/Haskell/FromCabal/PostProcess.hs
+++ b/src/Distribution/Nixpkgs/Haskell/FromCabal/PostProcess.hs
@@ -98,6 +98,7 @@ hooks =
   , ("mysql", set (libraryDepends . system . contains (pkg "mysql")) True)
   , ("network-attoparsec", set doCheck False) -- test suite requires network access
   , ("numeric-qq", set doCheck False) -- test suite doesn't finish even after 1+ days
+  , ("pango", set (libraryDepends . system . contains (bind "pkgs.gnome2.pango")) True)
   , ("pandoc", set jailbreak False) -- jailbreak-cabal break the build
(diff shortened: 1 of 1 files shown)


mdorman:
I still don't have a great story for testing these changes; please feel free to point out where I screwed up.

mdorman:
@peti I guess this doesn't apply as long as haskell-updates is based on 16.09 rather than master...

peti:
The `hackage-packages.nix` file in `release-16.09` has to use `pkgs.gnome.pango` -- which it does. In the `master` branch, however, you'll find that `pkgs.gnome2.pango` is used instead. I think this issue has been fixed a while ago by 9d0fb312eacdc94e5484f3bf469e497c3d189237 and 7e68a73d4c3d25493a0b5c3e732a9bc3081882eb.

peti:
Is there still a problem left to fix or can this PR be closed?

mdorman:
I don't know that there's anything left to fix.

Issue #90, opened 2014-09-01
Don't skip benchmarking sections


Fuuzetsu:
Currently `benchmark` section is just skipped. It'd be great if it could be accounted for.

expipiplus1:
Has there been any motion on this? Could #216 be used as inspiration?

TikhonJelvis:
I'm running into this problem now. Is there a reason the (relatively old) pull request fixing this hasn't been merged?

daig:
It looks like these days the dependencies are placed in `benchmarkHaskellDepends`, but are disabled by default and must be enabled via `withBenchmarkDepends = true`.
Since cabal 2.0 enables nix support through the `shell.nix`, and since it is against the [current design philosophy]() to add special support for `cabal2nix`, I think it would make sense to add `withBenchmarkDepends ? true` to the arguments generated by `cabal2nix --shell`. Otherwise there is no way to configure cabal for benchmarks without using from inside `nix-shell` directly, defeating the purpose of cabal's nix support.
In general since the `shell.nix` should be considered a dev environment, does it make sense to enable benchmark dependencies by default?

PR #262, opened 2017-02-27
Put benchmark dependencies in benchDepends

expipiplus1:
I'll open a PR for nixpkgs soon too


expipiplus1:
This is a partial solution to https://github.com/NixOS/cabal2nix/issues/90

support for benchDepends needs to be added in nixpgks.

PR #265, opened 2017-03-05
Fix and streamline `gobjectIntrospection`-based library handling

mdorman:
So there's four parts to this, each touching different groups of packages:

* `gstreamer` refactoring:

  1. Fixes name translation stuff to reflect changes recently made in master
  2. Refactors the existing helper functions into one parameterized function.

* `webkitgtk` refactoring:

  1. Fixes name translation stuff to minimize necessary overrides
  2. Has one function that handles that override, uses it where appropriate, removes everything else.

* `cairo` refactoring:

  1. Removes now-unnecessary `gobjectIntrospection` bits
  2. Change `gi-pangocairo` to use `giCairoPhaseOverrides` which is sufficient

* core `gdk`, `gtk`, etc. refactoring:

  1. Removes now-unnecessary `gobjectIntrospection` bits

There's few packages that use these modules, which makes it challenging to know whether you've broken something inadvertently, so I've adopted the following build tests:

* `tianbar`

  `tianbar` depends on a number of the affected packages:

    ```
    Dependency gi-gdk -any: using gi-gdk-3.0.11
    Dependency gi-gio -any: using gi-gio-2.0.11
    Dependency gi-glib -any: using gi-glib-2.0.11
    Dependency gi-gtk -any: using gi-gtk-3.0.11
    Dependency gi-webkit2 -any: using gi-webkit2-4.0.11
    ```

  Prior to these changes it is unbuildable, failing with:

    ```
    Setup: The pkg-config package 'webkit2gtk-4.0' is required but it could not be found.
    note: keeping build directory ‘/tmp/nix-build-tianbar-1.2.5.drv-3’
    builder for ‘/nix/store/3835k0qrw3nwywlb79q8q9n4yda4cpx4-tianbar-1.2.5.drv’ failed with exit code 1
    error: build of ‘/nix/store/3835k0qrw3nwywlb79q8q9n4yda4cpx4-tianbar-1.2.5.drv’ failed
    ```

  After this change, `tianbar` finds all its dependencies; however, it is still unbuildable---but that appears to be skew between interfaces rather than a problem with these changes:

    ```
    src/System/Tianbar/StrutProperties.hs:48:25: error:
        • Couldn't match expected type ‘Ptr Window’
                      with actual type ‘IO (Ptr b0)’
    ```

  I consider this to be progress.

* `vcsgui`

  `vcsgui` depends on a couple of the affected packages:

    ```
    Dependency gi-gtk >=3.0.6 && <3.1: using gi-gtk-3.0.11
    Dependency gi-gtk-hs >=0.3.0.0 && <0.4: using gi-gtk-hs-0.3.4.0
    ```

  Before this change, `vcsgui` built and runs, though it encounters an exception whose source isn't clear to me:

    ```
    ~/src/nixpkgs $ result/bin/vcsgui

    ** (vcsgui:16729): WARNING **: Error retrieving accessibility bus address: org.freedesktop.DBus.Error.ServiceUnknown: The name org.a11y.Bus was not provided by any .service files
    vcsgui: VCSException 1 "" "" "/home/forste/tmp/testvcs/hgreps/repo1" ["log","--style","xml","","--noninteractive"]
    ```

  Afterward, it builds and runs, though it encounters the same exception

* `reactive-banana-gi-gtk`

  `reactive-banana-gi-gtk` depends on `gi-gtk`:

    ```
    Dependency gi-gtk >=3.0.11 && <3.1: using gi-gtk-3.0.11
    ```

  Before this change, `reactive-banana-gi-gtk` built and passed its
  tests.  Afterward, it builds and passes its tests.  It does have
  some example code I have not yet tried to build.


mdorman:
@peti So, I have actually worked my way through most of the barriers to my alternative, and I'm pretty sure it's the better option---one thing I think is nice about it is that it is almost entirely about removing code from PostProcess.hs; almost every bit of the `Gir` stuff can just be dropped, and everything that depends on it that I've tested remains...well, as buildable as it was beforehand.

The challenge is that there aren't a lot of packages using the `gi-` libs, but I've been able to build some examples using `Gstreamer`, and after tweaking a few other things, I've got `tianbar` configuring, but not building (though it seems clear to me that it wasn't building before, either).

Anyway, I'm going to work on it some more, but I definitely think it's a better, more scalable solution if I can prove it works.

mdorman:
I think this is ready to go, though if you think of something else that you'd like me to test build or something, please let me know.

peti:
The updated package set is in https://github.com/peti/nixpkgs/commit/e8c18078b3f340cc5744bd40dcd8df98ac5a9c82. Builds on Hydra will run within 4 hours.

Issue #261, opened 2017-02-08
Support multi-project cabal repos


domenkozar:
```
$ cabal2nix https://github.com/plow-technologies/servant-auth.git
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   178  100   178    0     0     97      0  0:00:01  0:00:01 --:--:--   102
100 84066    0 84066    0     0  30299      0 --:--:--  0:00:02 --:--:--  147k
source archive /run/user/499/tmp.fZuzeDX8p2/servant-auth.git has unknown type
downloading ‘https://github.com/plow-technologies/servant-auth.git’... [0/0 KiB, 0.0  [30/0 KiB, 28.3 KiB/s]
*** parsing cabal file: https://github.com/plow-technologies/servant-auth.git: openBinaryFile: does not exist (No such file or directory)
Initialized empty Git repository in /run/user/499/git-checkout-tmp-Bjpua2a8/servant-auth/.git/
remote: Counting objects: 67, done.
remote: Compressing objects: 100% (45/45), done.
remote: Total 67 (delta 5), reused 39 (delta 1), pack-reused 0
Unpacking objects: 100% (67/67), done.
From https://github.com/plow-technologies/servant-auth
 * branch            HEAD       -> FETCH_HEAD
Switched to a new branch 'fetchgit'
removing `.git'...

git revision is 9b88328e9102fa5d9ec572aad8d9bbde4cef1f3f
path is /nix/store/9qcsazh52qamsskk9b88ifz800gbpqjl-servant-auth
git human-readable version is -- none --
Commit date is 2017-02-06 22:49:17 -0600
hash is 02fm3iaivq6kbc48kbs33fj7n3gjmhd7bpgc4ybs0b09z86krzm6
*** found zero or more than one cabal file. Exiting.
```

while trying to generate nix expression for servant-auth-server.

jmitchell:
Version 2.2 has been released and supports generating a nix expression when the input cabal file isn't in the repo's root directory. Example:

    cabal2nix --subpath servant-auth-server https://github.com/plow-technologies/servant-auth.git

peti:
The new version is available in Nixpkgs `master` at 1fdc34b73732715a74e926c4bee44c255df93360. Thank you very much, @jmitchell, for the fine implementation of this useful feature!

PR #271, opened 2017-05-01
Add --subpath to support repos with Cabal files in subdirectories.

jmitchell:
By default cabal2nix assumes the Cabal file will be found in the root
directory of the repo, but that's not always the case. Example use
case:

    cabal2nix --subpath servant-auth-server https://github.com/plow-technologies/servant-auth.git

The work is delegated to sourceRoot in nix's stdenv mkDerivation, and
manifests as the postUnpack attribute in the generated nix expression.


jmitchell:
This change is a partial solution to #261.

domenkozar:
@peti this allows for using `cabal2nix` with git repositores that don't have toplevel `*.cabal` but rather multiple projects. Example: https://github.com/serokell/engine.io

cabal2nix.cabal
```diff
@@ -16,39 +16,47 @@ author:         Peter Simons,
...
                 John Wiegley,
                 Philipp Hausmann,
                 Spencer Janssen,
+                Tom Hunger,
                 William Casarin,
+                koral,
                 Adam Vogt,
-                Bryan Gardiner,
+                Alexey Shmalko,
                 Corey O'Connor,
```
peti: What are all those modifications for? That doesn't seem right?

Issue #274, opened 2017-05-10
Generating an expression from a local file is broken since --subpath was added


peti:
`cabal2nix` can no longer convert a local Cabal file into a Nix expression:

~~~
$ cabal2nix --sha256=dontworry $(nix-build --no-out-link "<nixpkgs>" -A all-cabal-hashes)/parsec/3.0.0/parsec.cabal
*** parsing cabal file: /nix/store/spi1dcgsx2p4q526brdijkvkaglj1yhm-all-cabal-hashes-53fcf983669a3f0cdfd795fec28ecb40740a64ca-src/parsec/3.0.0/parsec.cabal/.: openBinaryFile: inappropriate type (Not a directory)
cabal2nix: nix-prefetch-zip: createProcess: runInteractiveProcess: exec: does not exist (No such file or directory)
~~~

Previous versions would simply emit a Nix file without trying to run any prefetch commands. According to `git bisect` the commit that introduced that issue is 85a20766f212626b081945118b21c8a1e51a35b6:

    $ git bisect log    
    git bisect start    
    # bad: [d2334f4e68a6a7d5a4a2b62dff0e62f81800a0b2] fromGenericPackageDescription: make sure the new subpath field is initialized    
    git bisect bad d2334f4e68a6a7d5a4a2b62dff0e62f81800a0b2    
    # good: [c75a2a15ca70b83826b925443e43ab41faed8649] Merge pull request #262 from expipiplus1/bench    
    git bisect good c75a2a15ca70b83826b925443e43ab41faed8649    
    # good: [9e55aabdc3d25be66c0c17a01904e557a77be08b] Bump version number to 2.1.1 for upcoming release.    
    git bisect good 9e55aabdc3d25be66c0c17a01904e557a77be08b    
    # good: [7e5a37f3813dda081a941b0b53158d59429e9f23] Merge pull request #269 from hamishmack/gtksouceview3    
    git bisect good 7e5a37f3813dda081a941b0b53158d59429e9f23    
    # good: [f4a9f3a656ca3de93744432c420b9cd1740a2910] package.yaml: update author list    
    git bisect good f4a9f3a656ca3de93744432c420b9cd1740a2910    
    # bad: [ff83d2d1913a1e4580dd0ff438b1833660854cb3] Cosmetic changes to the way postUnpack is rendered.    
    git bisect bad ff83d2d1913a1e4580dd0ff438b1833660854cb3    
[...]

PR #275, opened 2017-05-10
Fix for #274

jmitchell:
There's no fix yet, but there's a failing regression test we can use to verify a fix.

Submitting the PR now to see how Travis CI handles the regression test.


jmitchell:
@peti, the latest commit in this PR fixes #274. Still waiting on Travis to pass, but it succeeds on my machine.

The regression test verifies the generated nix expression is what we'd expect. Adding the test required some refactoring. Let me know if anything should be changed.

peti:
I cherry-picked the actual fix at 8abbd9fb3f6bd82bb3511f312aba018e28be36b9 so that we can have that bug fixed without needing to resolve all questions that arise due to the addition of the test suite first.

src/Cabal2nix.hs
```diff
@@ -0,0 +1,176 @@
...
+import Options.Applicative
+import Paths_cabal2nix ( version )
+import System.Environment ( getArgs )
+import System.IO ( hFlush, stderr, stdout )
+import qualified Text.PrettyPrint.ANSI.Leijen as P2 hiding ( (<$>), (<>) )
+import Text.PrettyPrint.HughesPJClass ( Doc, Pretty (..), text, vcat, hcat, semi )
+
+
+data Options = Options
+  { optSha256                    :: Maybe String
```
peti: Please do not make these kind of white-space changes while committing some other (unrelated) modifications. Cosmetic changes should have their own commits (or PRs) that contain nothing but the cosmetic so that these modifications don't clutter up the diff and effectively hide the substance of the patch. The diff of this PR is quite large, but the substance of the bug fix is the removal of a *single character*. It should be obvious from the diffs, IMHO, that this is the actual change the PR makes.

src/Cabal2nix.hs
```diff
@@ -0,0 +1,176 @@
...
+        )
+
+
+safeFlush :: IO () -> IO ()
+safeFlush io = bracket (return ()) (\() -> hFlush stdout >> hFlush stderr) $ \() -> io
+
+main :: IO ()
+main = mainWithArgs =<< getArgs
+
+mainWithArgs :: [String] -> IO ()
```
peti: Dito here with the refactoring. I like the introduction of the `safeFlush` function. But that change should have been made in a separate commit (or PR) --- not as part of something else.

src/Cabal2nix.hs
```diff
@@ -0,0 +1,176 @@
...
+mainWithArgs args = safeFlush $ do
+  Options {..} <- handleParseResult $ execParserPure defaultPrefs pinfo args
```
peti: Why did you replace the previous use of `execParser` with this construct? What are the advantages?

test/Main.hs
```diff
@@ -1,6 +1,44 @@
...
+        , "  libraryHaskellDepends = [ base bytestring mtl ];"
+        , "  homepage = \"http://www.cs.uu.nl/~daan/parsec.html\";"
+        , "  description = \"Monadic parser combinators\";"
+        , "  license = stdenv.lib.licenses.bsd3;"
+        , "}"
+        ]
+  ]
+
+
+allCabalHashes :: FilePath -> IO String
```
peti: This may be a bit much. The test does not require a 1,3 GByte git repository of Cabal files --- it just needs *one* Cabal file.

Also. note that this test fails when Nix runs it because there's no Nix available inside of the build sandbox. Neither does the sandbox have network access and the ability to download stuff from the Internet.

src/Cabal2nix.hs
jmitchell: I agree cosmetic changes should go in separate commits. The change was introduced because I had `stylish-haskell` running automatically, and after moving the code around I missed some of the changes it made. I'll be more careful about that next time.

src/Cabal2nix.hs
jmitchell: Got it :)

src/Cabal2nix.hs
jmitchell: `execParser` calls `getArgs` behind the scenes, but to facilitate end-to-end testing from Haskell I wanted a way to pass my own arguments `[String]`.

Another benefit of structuring the code this way is people can use features exposed to the CLI as a Haskell library. The alternative right now is using `System.Process`.

I can file a PR for this particular change if you agree it's worth having. Note it does require moving a lot of code into a non-`Main` module.

test/Main.hs
jmitchell: Agreed on all counts. I will create a simpler test case that uses a locally defined cabal file.

jmitchell:
> I'm not sure whether a Haskell-written test suite is the way to go for this kind of black box test. It feels a bit like overkill, i.e. in other projects a simple shell script would be used to accomplish this.

For the most part I agree, a shell script would be better. However, as nix support improves on Windows writing portable tests in Haskell will be simpler. Perhaps that's not such a big concern yet, though.

src/Cabal2nix.hs
peti: If we'd like to do regression testing of `cabal2nix` via the command-line interface, then having such a re-usable module makes sense. BTW, if you ever want to call a `main` function with specific arguments from a Haskell library, then `System.Environment.withArgs` can do it.

peti:
I suppose this PR is obsolete. Please let me know if I assumed incorrectly!

PR #278, opened 2017-06-06
Migrate cabal2nix's Main module to a library-exposed module.

jmitchell:
The motivation is to support calling cabal2nix from another Haskell
library using the same interface provided by the executable.

A similar change was suggested as part of #274, but this is a single-purpose proposal.


domenkozar:
cc @peti

PR #284, opened 2017-07-09
Expose derivation info from cabal2nix'

dmjio:
cc @peti , am working on generating nix builds from `GHCJS` stack files. This requires getting out the raw derivation information to construct a `hackage-packages.nix`-like file. 

This PR exposes `cabal2nix'`, which gives access to the underlying derivation, alleviating us from having to call `cabal2nix` as a subprocess, and also giving flexibility in regards to command line arguments (e.g. `--compiler ghcjs`).


dmjio:
@peti thanks

PR #289, opened 2017-08-03
Improve nix caching for local packages with subpaths

jmitchell:
Instead of generating this for local packages with subpaths

    src = ./.;
    postUnpack = "sourceRoot+=/dep";

this change generates

    src = ./dep;

The postUnpack strategy is still used for subpaths in remote
repositories.

### DIFF
--- a/src/Distribution/Nixpkgs/Haskell/Derivation.hs
+++ b/src/Distribution/Nixpkgs/Haskell/Derivation.hs
@@ -108,7 +108,7 @@ instance Pretty Derivation where
       [ attr "pname"   $ doubleQuotes $ disp (packageName _pkgid)
       , attr "version" $ doubleQuotes $ disp (packageVersion _pkgid)
       , sourceAttr _src
-      , onlyIf (_subpath /= ".") $ attr "postUnpack" postUnpack
+      , onlyIf (_subpath /= "." && isRemotePackage _src) $ attr "postUnpack" postUnpack
       , onlyIf (_revision > 0) $ attr "revision" $ doubleQuotes $ int _revision
       , onlyIf (not (null _editedCabalFile) && _revision > 0) $ attr "editedCabalFile" $ string _editedCabalFile
       , listattr "configureFlags" empty (map (show . show) renderedFlags)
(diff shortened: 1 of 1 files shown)


jmitchell:
## Remote repo with subpath

```
$ cabal2nix --subpath dep1 https://github.com/jmitchell/haskell-multi-package-demo1.git
...
{ mkDerivation, base, fetchgit, stdenv }:
mkDerivation {
  pname = "dep1";
  version = "0.1.0.0";
  src = fetchgit {
    url = "https://github.com/jmitchell/haskell-multi-package-demo1.git";
    sha256 = "0mxpn41qizy965n9ghjazisvx2ci0mmlq37iyqpji395qf0860s3";
    rev = "e3d9bd6d6066dab5222ce53fb7d234f28eafa2d5";
  };
  postUnpack = "sourceRoot+=/dep1; echo source root reset to $sourceRoot";
  libraryHaskellDepends = [ base ];
  license = stdenv.lib.licenses.bsd3;
}
```

## Local repo with subpath

```
$ cabal2nix --subpath dep1 .   # a local clone of the same repo
{ mkDerivation, base, stdenv }:
mkDerivation {
  pname = "dep1";
  version = "0.1.0.0";
  src = ./dep1;
  libraryHaskellDepends = [ base ];
  license = stdenv.lib.licenses.bsd3;
}
```

peti:
Why would anyone call `cabal2nix --subpath=foo .` instead of `cabal2nix foo` if all files required for the build reside in the sub-directory?

jmitchell:
Using `cabal2nix foo` would mean the resulting nix expression is only useful when the current directory is `foo`, but I use cabal2nix to generate many nix expressions that need to work from the same directory.

```
$ cabal2nix dep1
{ mkDerivation, base, stdenv }:
mkDerivation {
  pname = "dep1";
  version = "0.1.0.0";
  src = ./.;
  libraryHaskellDepends = [ base ];
  license = stdenv.lib.licenses.bsd3;
}
```

peti:
How about `cabal2nix $PWD/foo` then? That works everywhere.

jmitchell:
Ah, I was mixed up. My previous comment was based on the changes in this PR, but in `master` it produces the desired `src = ./dep1`. Thanks, @peti.

Issue #249, opened 2016-10-28
Support hpack


3noch:
If cabal2nix notices a `package.yaml` file, it could generate the `cabal` file first, right?

razvan-panda:
Any update on this? How hard would it be to implement? If no one offers I will do it.

3noch:
@razvan-panda No it's actually quite easy:

```nix
let

  hpack2cabal = name: src: pkgs.runCommand "hpack2cabal-${name}" {} ''
    ${pkgs.haskellPackages.hpack}/bin/hpack '${src}' - > "$out"
  '';

in haskellPackages.override {
  overrides = self: super: {
    callHpack2nix = name: src: self.callCabal2nix name (hpack2cabal name src);
  }
}
```

It would be easy to add variants that check for a `package.yaml` file first. Checking for a `cabal` file would not be good because many Hpack repos keep their cabal files around. I'm torn as to whether it is a good idea to do this by default in `callCabal2nix`, but certainly a `callHpack2nix` would be fine.

3noch:
I think best of all would be to add this directly to `cabal2nix` and get the behavior for everything out of the box.

razvan-panda:
@3noch Since you already did it - would you like to do a pull request? 👍

3noch:
@razvan-panda Sure. Haha...I've been meaning to. Thanks for the nudge.

razvan-panda:
@3noch Cheers! You are doing science's work! 👍

tomsmalley:
I have just implemented this in PR #304 so that hpack is used as a fallback (or can be forced by a command line flag).

tomsmalley:
@3noch the PR has now been merged!

ip1981:
@3noch did you mean
```
callCabal2nix = name: src: self.callCabal2nix name (hpack2cabal name src);
```
?

3noch:
No, I wasn't implementing it recursively. Does thta work?

ip1981:
I don't know :)

ip1981:
Guys, I'm trying to apply new cabal2nix to a package without `.cabal` file, and encountered another issue.

**2.4:** fetches sources, but can't find a `.cabal` file, ok.
```
# cabal2nix --version
cabal2nix 2.4.1

# cabal2nix https://github.com/commercialhaskell/stack >/dev/null
downloading ‘https://github.com/commercialhaskell/stack’... [0/0 KiB, 0.0 KiB/s]
[...]

PR #304, opened 2017-09-13
Try hpack when cabal file is missing (issue #249)

tomsmalley:
Does not take into account direct paths to the package.yaml file like `cabal2nix mypackage/package.yaml`.

The order of preference is cabal files > package.yaml files, i.e. if both are present we prefer to use the cabal file instead of regenerating it.

Needs documentation, is a short note in the usage menu enough?


tomsmalley:
Build stage fails without the cabal file present at configure time.
I could add a preConfigure stage that either:
* Runs hpack (cleaner but means running hpack twice + extra dependency in the nix expression)
* Writes the cabal2nix hpack output into the nix expression. This seems bad since the output would be messy, but it doesn't have the two issues of the above option (I'm not sure they are that bad though).

tomsmalley:
I implemented the first option above and also added a command line option to force using hpack in the configure process.

tomsmalley:
I think it's best to rely on the cabal file over the package.yaml file, allowing the user to override if they wish. Also added proper UTF8 encoding.

3noch:
@tomsmalley Fantastic! This really is the best option.

3noch:
Any progress on this?

tomsmalley:
This should be ready to merge, I do not know why travis is failing, seems to be unrelated to the PR?

src/Distribution/Nixpkgs/Haskell/Derivation.hs
```diff
@@ -59,6 +59,7 @@ data Derivation = MkDerivation
   , _editedCabalFile            :: String
   , _enableSeparateDataOutput   :: Bool
   , _metaSection                :: Meta
+  , _runHpack                   :: Bool
```
peti: The `Derivation` type contains only fields that can be passed as arguments to the generic Haskell builder, but the addition of `runHpack` breaks that rule. Therefore I don't think it's a good idea to add that field in here, especially since it's being mapped into a `preConfigure` hook, which is supposed to be used by *user code*, not by the proper Nixpkgs infrastructure. IMHO, it would be better to create an appropriate entry in `phaseOverrides` by some other means that does not rely on a `runHpack` attribute in `Derivation` itself. I understand it's a *command-line argument* to `cabal2nix`? So maybe that logic should be contained in `main`.

tomsmalley: I didn't realise that, I will amend the PR tomorrow.

tomsmalley:
@peti I think this makes more sense. Thanks for explaining!

src/Cabal2nix.hs
```diff
@@ -120,15 +121,21 @@ main :: IO ()
 main = bracket (return ()) (\() -> hFlush stdout >> hFlush stderr) $ \() ->
   cabal2nix =<< getArgs
 
+hpackOverrides :: Derivation -> Derivation
+hpackOverrides = over phaseOverrides (++ "preConfigure = \"${hpack}/bin/hpack;\";")
+               . set (libraryDepends . pkgconfig . contains (PP.pkg "hpack")) True
```
peti: Does `hpack` really belong into the pkgconfig dependencies? I would have placed it into `tools` (a.k.a. Build-Tools in Cabal). That should also put the tool into $PATH so that you can call it just as `hpack`.

src/Cabal2nix.hs
```diff
@@ -120,15 +121,21 @@ main :: IO ()
 main = bracket (return ()) (\() -> hFlush stdout >> hFlush stderr) $ \() ->
   cabal2nix =<< getArgs
 
+hpackOverrides :: Derivation -> Derivation
+hpackOverrides = over phaseOverrides (++ "preConfigure = \"${hpack}/bin/hpack;\";")
```
peti: The trailing `;` to end the command is probably not necessary.

src/Cabal2nix.hs
tomsmalley: Yes that makes more sense. I am quite new to nix, sorry.

tomsmalley:
@peti I have fixed the build and done the changes as requested. If there is anything else to change I am happy to do so.

PR #308, opened 2017-10-18
fix #307: hmatrix-special

### LINKED ISSUE #307: hmatrix-special doesn't build
I believe I have a fix, and can make a pull request.

Right now, hmatrix.cabal is overridden with the following:
```
preConfigure = "sed -i hmatrix.cabal -e 's@/usr/@/dont/hardcode/paths/@'"
```

My recommendation is to just delete the line in hmatrix.cabal instead:
```
preConfigure = "sed -i hmatrix.cabal -e '/\/usr\//D'"
```

With this change, `hmatrix.cabal` is still well-formed and compiles.  But we don't end up with a bad path that prevents dependent projects like hmatrix-special from compiling:

```
$ nix-build -A haskellPackages.hmatrix-special
...
impure path `/dont/hardcode/paths/lib/openblas/lib' used in link
collect2: error: ld returned 1 exit status
...
error: build of ‘/nix/store/xdc0swp1x8d7z1kfqxj3xvdcinfh8bp1-hmatrix-special-0.4.0.1.drv’ failed
```

The proposed fix allows both hmatrix and hmatrix-special to compile (I've tested it with overlays in nixpkgs).

shajra:
The prior code did a sed substitution of the dangerous /usr/ path in
hmatrix.cabal with a safer /dont/hardcode/path one.  That's fine for
compilation of hmatrix, but hmatrix-special (which depends on hmatrix)
then fails to compile because an impure (outside /nix/store) link is
detected.

Here's an illustration of the problem:

```
$ nix-build -A haskellPackages.hmatrix-special
...
impure path `/dont/hardcode/paths/lib/openblas/lib' used in link
collect2: error: ld returned 1 exit status
```

So this commit deletes the problematic reference in hmatrix.cabal,
rather than substitute it.  hmatrix.cabal is still well-formed and
hmatrix compiles.  But now hmatrix-special can compile too.

### DIFF
--- a/src/Distribution/Nixpkgs/Haskell/FromCabal/PostProcess.hs
+++ b/src/Distribution/Nixpkgs/Haskell/FromCabal/PostProcess.hs
@@ -75,7 +75,7 @@ hooks =
   , ("hfsevents", hfseventsOverrides)
   , ("HFuse", set phaseOverrides hfusePreConfigure)
   , ("hlibgit2 >= 0.18.0.14", set (testDepends . tool . contains (pkg "git")) True)
-  , ("hmatrix", set phaseOverrides "preConfigure = \"sed -i hmatrix.cabal -e 's@/usr/@/dont/hardcode/paths/@'\";")
+  , ("hmatrix", set phaseOverrides "preConfigure = \"sed -i hmatrix.cabal -e '/\\/usr\\//D'\";")
   , ("holy-project", set doCheck False)         -- attempts to access the network
   , ("hoogle", set testTarget "--test-option=--no-net")
   , ("hsignal < 0.2.7.4", set phaseOverrides "prePatch = \"rm -v Setup.lhs\";") -- https://github.com/amcphail/hsignal/issues/1


peti:
This patch does not compile.

shajra:
Hi @peti.  So sorry for the sloppy commit.  I was in a rush to move my work from the Nix expression I developed in Nixpkgs to the Haskell string in this project, I didn't exercise due diligence and committed in a string delimiting defect.

So I pushed a new version that compiles (I checked).  Since I already screwed up once, I'd prefer to run an end-to-end test.  I looked at the scripts in this repository, but I wasn't sure what the setup is to run them.  I'm assuming the right way to test this is to
1. checkout nixpkgs
1. checkout the "hackage" repository (not sure which one that is)
1. run your "update-nixpkgs.sh" script 
1. do a `nix-build -A haskellPackages.hmatrix-special` to see that it all worked end-to-end.

I'm happy to help with that test, but I need some pointers.

peti:
The easiest way to test that new build is to compile a new version of `cabal2nix`. Then generate build instructions:

    cabal2nix --shell cabal://hmatrix-special >shell.nix

And build the package:

    nix-build --no-out-link shell.nix

peti:
The generated build still fails:

~~~
setupCompilerEnvironmentPhase
Build with /nix/store/zip6rrw80vf1w0xs2958papzxb3kn48n-ghc-8.0.2.
unpacking sources
unpacking source archive /nix/store/vhq452spk4fcli4ic12i1qpk0n6kviz9-hmatrix-0.18.0.0.tar.gz
source root is hmatrix-0.18.0.0
setting SOURCE_DATE_EPOCH to timestamp 1478117796 of file hmatrix-0.18.0.0/THANKS.md
patching sources
compileBuildDriverPhase
setupCompileFlags: -package-db=/tmp/nix-build-hmatrix-0.18.0.0.drv-0/package.conf.d -j4 -threaded
[1 of 1] Compiling Main             ( Setup.lhs, /tmp/nix-build-hmatrix-0.18.0.0.drv-0/Main.o )
Linking Setup ...
configuring
sed: -e expression #1, char 3: unknown command: `u'
builder for ‘/nix/store/35k7gak8d80a9n6yi6ryfmjyfjbk96lk-hmatrix-0.18.0.0.drv’ failed with exit code 1
~~~

jackpot1136:
Thank you

Issue #313, opened 2017-11-26
2.6 regressed in performance


domenkozar:
2.6 takes ~2s for generating a package, while 2.5 took ~1s.

```
$ \time /nix/store/km80h0kh6m6qmjwhdq73i31vpnrc61cg-cabal2nix-2.5/bin/cabal2nix cabal://mtl
{ mkDerivation, base, stdenv, transformers }:
mkDerivation {
  pname = "mtl";
  version = "2.2.1";
  sha256 = "1icdbj2rshzn0m1zz5wa7v3xvkf6qw811p4s7jgqwvx1ydwrvrfa";
  revision = "1";
  editedCabalFile = "0fsa965g9h23mlfjzghmmhcb9dmaq8zpm374gby6iwgdx47q0njb";
  libraryHaskellDepends = [ base transformers ];
  homepage = "http://github.com/ekmett/mtl";
  description = "Monad classes, using functional dependencies";
  license = stdenv.lib.licenses.bsd3;
}
0.97user 0.21system 0:01.19elapsed 99%CPU (0avgtext+0avgdata 537108maxresident)k
0inputs+8outputs (0major+133288minor)pagefaults 0swaps

$ \time /nix/store/0g3x5r9paygsvfd9y8m1rsrri2amh990-cabal2nix-2.6/bin/cabal2nix cabal://mtl
{ mkDerivation, base, stdenv, transformers }:
mkDerivation {
  pname = "mtl";
  version = "2.2.1";
  sha256 = "cae59d79f3a16f8e9f3c9adc1010c7c6cdddc73e8a97ff4305f6439d855c8dc5";
  revision = "1";
  editedCabalFile = "0fsa965g9h23mlfjzghmmhcb9dmaq8zpm374gby6iwgdx47q0njb";
  libraryHaskellDepends = [ base transformers ];
  homepage = "http://github.com/ekmett/mtl";
  description = "Monad classes, using functional dependencies";
  license = stdenv.lib.licenses.bsd3;
}
1.89user 0.30system 0:02.20elapsed 100%CPU (0avgtext+0avgdata 593852maxresident)k
0inputs+0outputs (0major+147407minor)pagefaults 0swaps
```

domenkozar:
cc @jmitchell

domenkozar:
Bisecting leads to eef3b98ebd69d2f80b334e0d196a6cd2c9867b58

domenkozar:
[...]

PR #314, opened 2017-12-05
Optimization: allow HackageDB reuse

jmitchell:
As noted in #313, HackageDB deserialization is rather expensive. This
change allows Haskell projects to deserialize a HackageDB once, and
then make multiple calls to cabal2nix with that DB.

An experimental result with stack2nix shows over a 6x speed
improvement with pandoc and its dependency closure.


domenkozar:
cc @peti

Issue #315, opened 2017-12-12
Internal Libraries in Cabal 2.+


ts468:
Cabal 2.0 and later support “internal libraries” [1]. Internal libraries are named library sections in a `.cabal`-file that can be referenced in other build-depends like any other, external package, but are resolved internally by cabal.

Cabal2nix seems to generate external dependencies for internal libraries as well. Did somebody already look into extending cabal2nix to filter out internal libraries before generating the external dependencies?

[1] https://www.haskell.org/cabal/users-guide/developing-packages.html#library

tomsmalley:
See PR #316

ts468:
@tomsmalley @peti Thank you so much for your efforts!!!

PR #316, opened 2017-12-17
Support for Cabal 2.x internal library (issue #315)

tomsmalley:
- Filter the dependency set using the subLibraries names
- Add the subLibraries deps to the derivation deps

I **think** the second is also required to provide the right dependencies for `nix-shell`.

### DIFF
--- a/src/Distribution/Nixpkgs/Haskell/FromCabal.hs
+++ b/src/Distribution/Nixpkgs/Haskell/FromCabal.hs
@@ -129,9 +129,12 @@ fromPackageDescription haskellResolver nixpkgsResolver missingDeps flags (Packag
     internalLibNames = fmap unqualComponentNameToPackageName . catMaybes
                      $ libName <$> subLibraries
 
+    internalLibDeps :: [Dependency]
+    internalLibDeps = targetBuildDepends . libBuildInfo =<< subLibraries
+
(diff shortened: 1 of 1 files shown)


src/Distribution/Nixpkgs/Haskell/FromCabal.hs
```diff
@@ -124,9 +125,16 @@ fromPackageDescription haskellResolver nixpkgsResolver missingDeps flags (Packag
     resolveInHackageThenNixpkgs i | haskellResolver (Dependency (mkPackageName (i^.ident)) anyVersion) = resolveInHackage i
                                   | otherwise = resolveInNixpkgs i
 
+    internalLibNames :: [PackageName]
+    internalLibNames = fmap unqualComponentNameToPackageName . catMaybes
+                     $ libName <$> subLibraries
+
+    internalLibDeps :: [Dependency]
+    internalLibDeps = targetBuildDepends . libBuildInfo =<< subLibraries
```
peti: How about the other dependencies recorded in a `BuildInfo` record? I assume that internal libraries can depend on system libraries and build tools, too?

tomsmalley:
Ah sorry I missed the first message. Yes I think you are right there, I will make adjustments tomorrow (hopefully). It’s something I forgot about.

PR #318, opened 2017-12-27
unknownPackageError: clarify what package fails

domenkozar:


domenkozar:
Thanks @peti for swift responses!

PR #321, opened 2018-01-01
Make HackageDB lazy to avoid unnecessary parsing

### LINKED ISSUE #319: 2.7.1 breaks callCabal2nix
I was checking some stuff against `nixpkgs/master` and `callCabal2nix` seems to always fail with

```
cabal2nix: NoHackageTarballFound
```

I am testing on macOS.  It seems to happen for anything using `callCabal2nix`.  Changing the version of `cabal2nix` to 2.7 (and updating the hash) in `hackage-packages.nix` seems to work around the issue.

domenkozar:
I don't like this solution too much, but it does the job. 

Fixes #319


peti:
https://hackage.haskell.org/package/cabal2nix-2.7.2

PR #322, opened 2018-01-05
Use identify function over pkgs set for nixpkgs resolving

domenkozar:
Currently cabal2nix for package `zlib` generates `zlib = null` binding, I think `zlib = pkgs.zlib` makes more sense for a default.

### DIFF
--- a/src/Cabal2nix.hs
+++ b/src/Cabal2nix.hs
@@ -161,7 +161,7 @@ processPackage Options{..} pkg = do
       withHpackOverrides = if pkgRanHpack pkg then hpackOverrides else id
       deriv :: Derivation
       deriv = withHpackOverrides $ fromGenericPackageDescription (const True)
-                                            (\i -> Just (binding # (i, path # [i])))
+                                            (\i -> Just (binding # (i, path # ["pkgs", i])))
                                             optSystem
                                             (unknownCompilerInfo optCompiler NoAbiTag)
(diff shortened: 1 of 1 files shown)


domenkozar:
Hmm, this doesn't work well with X11 package.

domenkozar:
It generates:

```
         }) {inherit (pkgs) libX11; inherit (pkgs) libXext; 
             inherit (pkgs.xorg) libXext; inherit (pkgs.xorg) libXinerama; 
             inherit (pkgs) libXrandr; inherit (pkgs.xorg) libXrender;};
```

domenkozar:
OK, this is best we can do for now.

peti:
Won't this generate evaluation errors for package names that don't exist?

domenkozar:
Yes. I plan to review how resolving works and find a better way to do this. I'm fine if you reject this, but that's what stack2nix does atm.

peti:
Well, I'm reluctant to have `cabal2nix` generate expression that fail to evaluate.

peti:
I would like clean up the code that resolves system dependencies, too, and make it available in https://github.com/peti/distribution-nixpkgs in a version that `hackage2nix`, `cabal2nix`, and other tools can re-use easily. It's not rocket science, but it's a somewhat ugly task because one has to employ heuristics that guess the best choice rather than being able to perform a reliable lookup.

PR #323, opened 2018-01-06
X11: add libXScrnSaver to system library depends

rvl:
Fixes NixOS/nixpkgs#33363.

I'm not yet able to run hackage2nix to check this change (wrong cabal version), but pretty sure it will work.


rvl:
Have just checked and the result of this change still successfully evaluates.

PR #328, opened 2018-01-23
Fix build for hpack-0.22.0.

tomsmalley:
This should probably wait for a reply from the hpack author for clarification of the extra filepath argument, but I think it can be ignored. This commit works for hpack-0.21 and hpack-0.22.

Can remove the macros if you'd rather just drop support for <0.22.


peti:
Personally, I'd happy to require hpack >= 0.22 and to just ignore older versions.

PR #339, opened 2018-03-19
Add `Windows` to platforms.

angerman:
That way we can use `cabal2nix` when cross compiling to windows as well.


Ericson2314:
CC @kmicklas. I think this is fine for now even if we'll replace the whole function?

Issue #333, opened 2018-02-15
don't depend on LANG=utf-8 being set for interpreting cabal files' encoding properly


sol:
Hey!  This probably obvious to everybody with minimal Nix experience, but I lack that basic experience. So here is what I try:

```
$ nix-shell -p cabal2nix --pure --run 'cabal2nix .'
*** parsing cabal file: ././tinc.cabal: hGetContents: invalid argument (invalid byte sequence)
cabal2nix: nix-prefetch-zip: createProcess: runInteractiveProcess: exec: does not exist (No such file or directory)
```

If I run this without `--pure` or install `cabal2nix` into my profile then it works, even though there is no `nix-prefetch-zip` on my `PATH`.  What additional package do I need to pass to `nix-env` to make this work?

sol:
Indeed, the following works:

```
nix-shell -p cabal2nix --pure --run 'LC_ALL=en_US.UTF-8 cabal2nix .'
```

Note that `.cabal`-files are UTF-8 per specification.  So at this point I would label this a bug, either in `cabal2nix` or one of it's dependencies (`nix-prefetch-zip`?!).

sol:
For completeness: https://www.haskell.org/cabal/users-guide/developing-packages.html#creating-a-package

> a Unicode UTF-8 text file containing a package description

peti:
This is a bug in `readGenericPackageDescription` from `Cabal`, which depends on the system locale to interpret Cabal files as utf-8-encoded. As far as I can tell, this issue is fixed in the upcoming 2.2.x release.

PR #341, opened 2018-03-19
Do not fail on .cabal files that contain Unicode (fixes #340)

### LINKED ISSUE #340: cabal2nix fails on .cabal files that contain Unicode characters
This regression was introduced with 9ef6995.  See the discussion at https://github.com/NixOS/cabal2nix/pull/334#issuecomment-372872783.

sol:


src/Distribution/Nixpkgs/Haskell/PackageSourceSpec.hs
```diff
@@ -193,18 +194,21 @@ cabalFromDirectory False dir = do
...
-      let hpackOutput = Hpack.renderPackage [] pkg'
-          hash = printSHA256 $ digestString (digestByName "sha256") hpackOutput
-      case runParseGenericPackageDescription "<hpack output>" (BS.pack hpackOutput) of
+      let hpackOutput = encodeUtf8 $ Hpack.renderPackage [] pkg'
+          hash = printSHA256 $ digest (digestByName "sha256") hpackOutput
+      case runParseGenericPackageDescription "<hpack output>" hpackOutput of
         Left msg -> liftIO $ do
           hPutStrLn stderr "*** hpack output:"
-          hPutStrLn stderr hpackOutput
+          BS.hPutStrLn stderr hpackOutput
```
sol: Outputting the plain `ByteString` here is not awesome and will result in gibberish if the locale is not set to UTF8.  However, outputting the `String` as it was before results in an exception if the locale is e.g. `c` and the output contain Unicode code points `> 127`.

Not sure if I like the current code, but also don't know any better option.  We could do a `show` on `hpackOutput` first and then print this to `stderr`, but as the output is potentially very large that would not be awesome neither.

Any suggestions welcome!

src/Distribution/Nixpkgs/Haskell/PackageSourceSpec.hs
```diff
@@ -216,11 +220,9 @@ cabalFromFile failHard file =
   -- that do not represent valid characters. To catch that exception, we need to
   -- wrap the whole block in `catchIO`.
   MaybeT $ handleIO (\err -> Nothing <$ hPutStrLn stderr ("*** parsing cabal file: " ++ show err)) $ do
-    buf <- withFile file ReadMode $ \ h -> do
-      hSetEncoding h utf8
-      hGetContents h >>= Exception.evaluate . force
-    let hash = printSHA256 (digestString (digestByName "sha256") buf)
```
sol: `digestString` is problematic, as it depends on the locale.

sol:
(fix for #340)

sol:
I was looking at adding a test case for this.  But doing so is not trivial, as AFAICT `Distribution.Nixpkgs.Haskell.PackageSourceSpec` is currently not penetrated by the golden tests.

@peti I assume that addressing your comment from https://github.com/NixOS/cabal2nix/commit/66cb4b641a29d3daf700b02c7706a79a68e9bf1d alone would not detect #333 nor #340.  In addition we would need to make sure that `Distribution.Nixpkgs.Haskell.PackageSourceSpec` is penetrated by the tests.

peti:
Personally, I'd prefer a path where `Distribution.Nixpkgs.Haskell.PackageSourceSpec` is removed from the source code and never to be seen again. That code is a bona fide testing and maintenance nightmare. I'll do some drastic refactoring in the future ...

Anyhow, thank you for the quick fix!

PR #344, opened 2018-04-01
Support all architectures and OSs that GHC and Cabal support

kmicklas:
This replicates GHC's normalization logic so that Cabal's built in parsers can
be used. This should ensure maximum compatibility with how the Cabal file will
actually be evaluated during the build.

Additionally, this adds support for passing in the full LLVM style platform
triple/quadruple (e.g. x86_64-unknown-linux-gnu) instead of just the short Nix
style (x86_64-linux). This will make it possible for the ABI to treated as part
of the OS where applicable (e.g. linux-android). We can then change
`make-package-set.nix` to use `hostPlatform.config` instead of
`hostPlatform.system` to get the full information.

### DIFF
--- a/src/Cabal2nix.hs
+++ b/src/Cabal2nix.hs
@@ -1,5 +1,6 @@
-{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards #-}
 
 module Cabal2nix
   ( main, cabal2nix, cabal2nix', cabal2nixWithDB
@@ -9,7 +10,9 @@ module Cabal2nix
 import Control.Exception ( bracket )
 import Control.Lens
 import Control.Monad ( when )
-import Data.Maybe ( fromMaybe, isJust )
+import Data.List
+import Data.List.Split
+import Data.Maybe ( fromMaybe, isJust, listToMaybe )
 import Data.Monoid ( (<>) )
 import qualified Data.Set as Set
 import Data.String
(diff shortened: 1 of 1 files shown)


src/Cabal2nix.hs
```diff
@@ -99,19 +102,51 @@ readP p = eitherReader $ \s -> case [ r' | (r',"") <- P.readP_to_S p s ] of
...
-                                    , P.string "mipsel" >> return Mips
-                                    ]
-                   _ <- P.char '-'
-                   os <- P.choice [ P.string "linux" >> return Linux, P.string "darwin" >> return OSX
-                                  , P.string "windows" >> return Windows]
-                   return (Platform arch os)
+-- | Replicate the normalization performed by GHC_CONVERT_CPU in GHC's aclocal.m4
+-- since the output of that is what Cabal parses.
+ghcConvertArch :: String -> String
+ghcConvertArch = \case
```
peti: Is the use of LambdaCase here really necessary? It feels like this code could easily have been written with normal pattern matching and it would look almost exactly the same.

src/Cabal2nix.hs
```diff
@@ -99,19 +102,51 @@ readP p = eitherReader $ \s -> case [ r' | (r',"") <- P.readP_to_S p s ] of
...
+  "i486"  -> "i386"
+  "i586"  -> "i386"
+  "i686"  -> "i386"
+  "amd64" -> "x86_64"
+  arch -> fromMaybe arch $ listToMaybe $
```
peti: hlint complains that the second $ is superfluous.

src/Cabal2nix.hs
```diff
@@ -99,19 +102,51 @@ readP p = eitherReader $ \s -> case [ r' | (r',"") <- P.readP_to_S p s ] of
...
+
+-- | Replicate the normalization performed by GHC_CONVERT_OS in GHC's aclocal.m4
+-- since the output of that is what Cabal parses.
+ghcConvertOS :: String -> String
+ghcConvertOS = \case
+  "watchos"       -> "ios"
+  "tvos"          -> "ios"
+  "linux-android" -> "linux-android"
+  os | "linux-" `isPrefixOf` os -> "linux"
+  os -> fromMaybe os $ listToMaybe $
```
peti: dito

src/Cabal2nix.hs
kmicklas: I generally prefer to not duplicate identifiers unnecessarily - especially with pattern guards that apply to the thing being cased on: `case foo of { foo' | foo' == bar -> ... }` or `case foo of { _ | foo == bar -> ... }`?

Happy to change the style if you want it the other way though. :)

peti: I would prefer the conventional style, yes. I believe that it can be written quite straight-forwardly without duplicating any identifiers (although I haven't actually tried it so I might be wrong).

kmicklas: Done with use of an underscore.

PR #351, opened 2018-05-28
fetchOrFromDB: Show which source failed when calling `fail`

nh2:
Note I haven't built this because I could find neither instructions against which nixpkgs I should best build this nor a `stack.yaml` so I'll just leave it here relying on your goodwill to typecheck it for me @peti :-)


peti:
Running `cabal new-build` and/or `cabal new-test` is the way I use to build this project. :-)

Anyhow, thank you for the contribution.

PR #352, opened 2018-05-28
hackage2nix: clear homepage

ryantm:
The hackage page is always there, and supports a TLS connection.

### DIFF
--- a/hackage2nix/Main.hs
+++ b/hackage2nix/Main.hs
@@ -27,7 +27,7 @@ import Distribution.Nixpkgs.Haskell.FromCabal.Flags
 import Distribution.Nixpkgs.Meta
 import Distribution.Nixpkgs.PackageMap
 import Distribution.Package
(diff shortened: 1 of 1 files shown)


hackage2nix/Main.hs
```diff
@@ -153,6 +153,7 @@ main = do
                   & metaSection.hydraPlatforms %~ (`Set.difference` Map.findWithDefault Set.empty name (dontDistributePackages config))
                   & metaSection.maintainers .~ Map.findWithDefault Set.empty name globalPackageMaintainers
                   & metaSection.hydraPlatforms %~ (if isInDefaultPackageSet then id else const Set.empty)
+                  & metaSection.homepage .~ ("https://hackage.haskell.org/package/" ++ display pkgId)
```
peti: Please set `homepage` to an empty string `""` instead to suppress the `meta.homepage` outout altogether. That way, we can let the generic builder construct the appropriate URL from its default settings rather than repeating an almost identical Hackage URL several thousand times in the generated file.

PR #353, opened 2018-05-31
Fix dbusmenu-gtk3

puffnfresh:
It also has a dependency on gtk3 which is not being picked up.


puffnfresh:
I've submitted a PR to gi-dbusmenu if cabal2nix is not the appropriate place for this change.

PR #360, opened 2018-07-17
If both package.yaml and cabal exist, prefer package.yaml.

nh2:
This is because in a nomal workflow, you would generate the .cabal
every now and then from the package.yaml file.

If you forget to do this generation and you have an old .cabal file
in your tree (common because for users of hpack, .cabal files are
usually in .gitignore), then until now nix would pick up the outdated
cabal file and you would accidentally be running against outdated cabal
contents.

So in this commit we change the logic from

  "if *.cabal doens't exist, check if package.yaml exist, if so use hpack"

to

  "if package.ymal exists, use hpack, otherwise use *.cabal"


qrilka:
I think this PR made `cabal2nix` CLI flags inconsistent, in `cabal2nix --help` I see the following:
```
  --hpack                  run hpack before configuring this package (only
                           non-hackage packages
```
And yet even if I don't specify this flag `hpack` gets uses if there is `package.yaml`.
So in the end `hpack` gets run not depending on whether was `--hpack` flag specified or not (though behavior differs a bit when there is no `package.yaml`).
As a way around that situation I plan to use `data UseHpack = ForcedHpack | DefaultHpack | NoHpack` instead of just a boolean and add `--no-hpack` flag to make use of the 3rd option

PR #368, opened 2018-08-09
Add override for the vte system library

cdepillabout:
Add an override to the vte system library so that `gnome3.vte` is used instead of `gnome2.vte`.
This is similar to how gtk and gtksourceview work.

This is necessary so that Haskell packages that depend on VTE get setup to use VTE from Gnome3 instead of Gnome2.

This was suggested in https://github.com/NixOS/nixpkgs/pull/44529#discussion_r208072760.

An overview of why this is needed for the `gi-vte` package can be found [here](https://github.com/NixOS/nixpkgs/pull/44529#issuecomment-411624086).

Ideally I'd like to get this PR in to cabal2nix in order to get `gi-vte` working in time for the 18.09 release.


src/Distribution/Nixpkgs/Haskell/FromCabal.hs
```diff
@@ -139,6 +139,7 @@ fromPackageDescription haskellResolver nixpkgsResolver missingDeps flags Package
       | i == "gtk2"                      = binding # (i, path # ["pkgs","gnome2","gtk"])
       | i == "gtk3"                      = binding # (i, path # ["pkgs","gnome3","gtk"])
       | i == "gtksourceview3"            = binding # (i, path # ["pkgs","gnome3","gtksourceview"])
+      | i == "vte"                       = binding # (i, path # ["pkgs","gnome3","vte"])
```
peti: This change will break all packages that assume `vte` to come from `gnome2`. Note that, unlike the example above it, this entry has no "3" baked into its name and is therefore ambiguous.

cdepillabout: Good catch @peti.  I just created commit 5fab8e6 that I think will fix this.  It specifically checked for the `vte-2.91` library (instead of just any version of `vte`).

You can see that the `gi-vte` Haskell package is specifically depending on `vte-2.91` from Gnome 3:

http://hackage.haskell.org/package/gi-vte-2.91.18/src/gi-vte.cabal

However, the older `vte` Haskell package is just depending on the `vte` library from Gnome 2:

https://github.com/gtk2hs/vte/blob/edd23698501af524610153121eff15a03d9c41c3/vte.cabal-renamed#L72

With commit 5fab8e6, both the `gi-vte` and `vte` Haskell packages should get created correctly by hackage2nix.

How does this look?

jtojnar:
`vte-2.91` is pkg-config name. If it gets bumped, we will package it as a new expression (vte4).

peti:
It is categorically impossible to release a bugfix for `vte-2.91` that has a higher version number yet still goes into the gnome3 hierarchy?

jtojnar:
If the `.pc` name file was changed before the release GNOME 4, we would create a new package `gnome3.vte4`, or more likely move all VTE to top-level as has been done with GtkSourceView:

https://github.com/NixOS/nixpkgs/commits/a9dbcdb6269a68fdce7b229f9ce98f8b241cd60b/pkgs/development/libraries/gtksourceview

Either way, an attribute should have stable `.pc` name.

cdepillabout:
Hi @peti, any chance to take a look at this?  I believe this PR should be good to be merged as-is.

src/Distribution/Nixpkgs/Haskell/FromCabal/Name.hs
```diff
@@ -147,6 +147,7 @@ libNixName "taglib_c"                           = return "taglib"
 libNixName "tensorflow"                         = return "libtensorflow"
 libNixName "udev"                               = return "systemd";
 libNixName "uuid"                               = return "libossp_uuid";
+libNixName "vte-2.91"                           = return "vte-2.91"
```
basvandijk: This will result in a Nix expression with invalid syntax since a `.` is not allowed in a Nix identifier:

```
$ nix eval '(let vte-2.91 = "foo"; in vte-2.91)'
error: syntax error, unexpected FLOAT, expecting '.' or '=', at (string):1:11
```

basvandijk: Introducing a fake name like "vte_291" might work. So:

```haskell
libNixName "vte-2.91" = return "vte_291"
```

```haskell
    resolveInNixpkgs i
      ...
      | i == "vte_291" = binding # (i, path # ["pkgs","gnome3","vte"])
```

I'm not sure this is a hack but I think it might work.

jtojnar: Would moving vte to nixpkgs top-level help? I think we will want to do that eventually anyway.

cdepillabout: @basvandijk I was under the impression that it would be okay because I was overriding it in both [`resolveInNixPkgs`](https://github.com/NixOS/cabal2nix/pull/368/files#diff-500289182850937444c5f7f86d609268R137) and [`libNixName`](https://github.com/NixOS/cabal2nix/pull/368/files#diff-cd05d955733f4bc3562fec4757d99eaeR150).

But I'll go ahead and change it just in case!

shlevy: FYI, you can have dots in attribute names if they're quoted: `{ "foo.bar" = null; }`

cdepillabout:
@basvandijk @jtojnar @shlevy  Thanks for reviewing this PR.

@basvandijk I've made a change like you suggested in commit 5955759.

The fake name `vte_291` has been introduced.  It is matched on in the `resolveInNixpkgs` function.

basvandijk:
@cdepillabout thank you!

cdepillabout:
@basvandijk Thanks for reviewing and merging it in!

PR #370, opened 2018-08-10
Add --no-submodules option to suppress fetching submodules (e.g. git).

kquick:
There are times when submodule fetching is unnecessary (e.g. the submodules will be built as separate nix packages), so it would be nice to suppress this for git repos with extensive or recursive submodules.  This patch leaves the default behavior to fetch the submodules, but adds an option to disable that.

    $ diff <(time cabal2nix https://github.com/GaloisInc/macaw-semmc --subpath macaw-semmc 2> /dev/null) \
           <(time cabal2nix https://github.com/GaloisInc/macaw-semmc --subpath macaw-semmc --no-submodules 2> /dev/null)
    
    real    0m2.169s
    user    0m0.105s
    sys     0m0.037s

    real    1m16.636s
    user    0m6.695s
    sys     0m1.449s
    11c11
    <     sha256 = "16gmdhxn92qxpqlx2p2bd4cg1zpjsv5qa8hf4riymsifndrfcnvx";
    ---
    >     sha256 = "0yc5y9vcvkri2cj5mma37jca07srlm8h089pqrq9zh9g63aa1b9w";
    $

### DIFF
--- a/src/Cabal2nix.hs
+++ b/src/Cabal2nix.hs
@@ -91,7 +91,7 @@ options = Options
           <*> optional (option utcTimeReader (long "hackage-snapshot" <> help "hackage snapshot time, ISO format"))
           <*> pure (\i -> Just (binding # (i, path # [ident # "pkgs", i])))
           <*> strArgument (metavar "URI")
-          <*> flag True False (long "no-submodules" <> help "don't fetch git submodules")
+          <*> flag True False (long "dont-fetch-submodules" <> help "do not fetch git submodules from git sources")
 
 -- | A parser for the date. Hackage updates happen maybe once or twice a month.
 -- Example: parseTime defaultTimeLocale "%FT%T%QZ" "2017-11-20T12:18:35Z" :: Maybe UTCTime
--- a/src/Cabal2nix.hs
+++ b/src/Cabal2nix.hs
@@ -64,6 +64,7 @@ data Options = Options
   , optHackageSnapshot :: Maybe UTCTime
   , optNixpkgsIdentifier :: NixpkgsResolver
   , optUrl :: String
+  , optFetchSubmodules :: Bool
   }
 
 options :: Parser Options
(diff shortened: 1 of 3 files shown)


kquick:
The travis failure is an `hlint` suggestion which results in more obscure usage, IMHO, so I will change this if you would like, but I think it's clearer in the current form.

src/Distribution/Nixpkgs/Fetch.hs
```diff
@@ -72,16 +72,18 @@ fromDerivationSource :: DerivationSource -> Source
...
                                                 -- This is required, because we cannot always check if a download succeeded otherwise.
       -> Source                                 -- ^ The source to fetch from.
       -> IO (Maybe (DerivationSource, a))       -- ^ The derivation source and the result of the processing function. Returns Nothing if the download failed.
-fetch f = runMaybeT . fetchers where
+fetch s f = runMaybeT . fetchers where
   fetchers :: Source -> MaybeT IO (DerivationSource, a)
   fetchers source = msum . (fetchLocal source :) $ map (\fetcher -> fetchWith fetcher source >>= process)
     [ (False, "url", [])
-    , (True, "git", ["--fetch-submodules"])
+    , (True, "git", if s then ["--fetch-submodules"] else [])
```
basvandijk: IMHO if you name `s` a bit more descriptively like `optSubmodules` which is also consistent with the call site then the hlint suggestion of using: 
```haskell
[ "--fetch-submodules" | optSubmodules ]
``` 
is quite clear and understandable. It's a bit like the `optional` function in Nix.

kquick: That's a good suggestion, thank you.

src/Cabal2nix.hs
```diff
@@ -90,6 +91,7 @@ options = Options
           <*> optional (option utcTimeReader (long "hackage-snapshot" <> help "hackage snapshot time, ISO format"))
           <*> pure (\i -> Just (binding # (i, path # [ident # "pkgs", i])))
           <*> strArgument (metavar "URI")
+          <*> flag True False (long "no-submodules" <> help "don't fetch git submodules")
```
basvandijk: The last nitpick: what about naming the flag
`--dont-fetch-submodules`?
Since that name is more consistent with the flag we disable.

kquick: I personally dislike the "dont" terminology which drops the apostrophe, but `--do-not-fetch-submodules` gets even wordier.  In this case though I accede to the utility of consistency in the terminology.

basvandijk:
Thank you!

PR #371, opened 2018-08-15
Output fetchSubmodules setting in src directive

kquick:
This was overlooked in my previous (accepted) patch which adds the `--dont-fetch-submodules` command-line flag: the generated nix specification for `src` should reflect the setting of that flag so that the corresponding `fetchgit` operation behaves similarly.  This has no effect for `src` directives that do not use `fetchgit`.


kquick:
@basvandijk: sorry, but I missed this functionality in the previous patch which added the `--dont-fetch-submodules`.  I've tested the results, and it all should be working now with the addition of this PR.

PR #375, opened 2018-09-05
Allows for package definition in package.dhall

### LINKED ISSUE #374: Where hpack goes so does hpack-dhall.
In the same way that we have an `--hpack` option for taking the package definition from `package.yaml` with [hpack](https://github.com/sol/hpack) can we also have an `--hpack-dhall` option for taking the package definition from `package.dhall` with [hpack-dhall](https://github.com/sol/hpack-dhall)?
```

> cabal2nix --help
cabal2nix converts Cabal files into build instructions for Nix.

Available options:
--hpack                  run hpack before configuring this package (only
                           non-hackage packages)
```

philderbeast:
Fixes #374.


nh2:
This seems tough.

The `cabal` file systax was not expressive enough for some users, so alternatives sprung up. As a consequence, one cannot any longer analyse git checkouts directly from git checkouts (as no cabal file exists and is to be generated).

If `cabal2nix` restricted itself to only `cabal` files as input, how would we invoke such preprocessors? Also considering other downstream tools, such as `stack2nix`.

philderbeast:
Being able to point cabal2nix at hackage and code repos is very useful.

With the growing number of alternative ways to describe packages, each as far as I know convertible to cabal package descriptions, perhaps cabal2nix should restrict itself to `*.cabal` files in the next major release?

* From `package.yaml` to `<name>.cabal` with [hpack](https://github.com/sol/hpack).
* From `package.dhall` to `<name>.cabal` with [hpack-dhall](https://github.com/BlockScope/hpack-dhall).
* From `<name>.dhall` to `<name>.cabal` with [dhall-to-cabal](https://github.com/dhall-lang/dhall-to-cabal).

As a package maintainer I would prefer a cabal2nix blinkered to only see `*.cabal` files. I don't want it to run tooling on some upstream file and ignore a `*.cabal` file.

peti:
I am sorry, but this pull request goes into the opposite direction of where I want to go. I want to get support for `hpack` and `dhall` and stuff out of cabal2nix, not into it. I'll introduce a more modular architecture in the 3.x version that will allow you to implement these features as plugins without touching cabal2nix itself.

PR #380, opened 2018-10-10
Add `isSimpleBuild` to generated derivations

duog:
This will allow generic-builder.nix to use a shared Setup executable for all
packages that have `isSimpleBuild = true`. The default must be false, for
safety, even though almost all packages are simple.


peti:
Ping?

PR #393, opened 2018-12-04
Cabal revisions

qrilka:
Our final target is to use this in `stack2nix`, this PR depends on https://github.com/commercialhaskell/stack/pull/4412 so I guess we'll need `hackage-db` bounds updated when it will get merged


peti:
I don't think I  follow ... why does this PR depend on `stack`?

qrilka:
Oh, dummy copy-paste, sorry, linke corrected

qrilka:
@peti in ced478e8ee78e0d1bd75e86f0073cc1f30ffa515 I've added support for Stack's size part after SHA (there it's used as a measure against overflow attacks). Also in SHA256 is assumed to be in standard hexadecimal notation (Nix's base32 is used internally).
Please let me know what you think about this. Also it would be great to get some understanding of the situation with https://github.com/peti/hackage-db/pull/9

PR #402, opened 2019-01-29
Put file src attribute in quotes to allow more paths

qrilka:
Nix paths allow only limited set of characters thus for example paths with "@" of paths with non-latin characters give derivation with Nix fails to parse. Quoting resolves this problem.

### DIFF
--- a/src/Distribution/Nixpkgs/Fetch.hs
+++ b/src/Distribution/Nixpkgs/Fetch.hs
@@ -78,7 +78,7 @@ instance PP.Pretty DerivationSource where
     let isHackagePackage = "mirror://hackage/" `L.isPrefixOf` derivUrl
         fetched = derivKind /= ""
     in if isHackagePackage then attr "sha256" $ string derivHash
-       else if not fetched then attr "src" $ text derivUrl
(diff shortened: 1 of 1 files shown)


qrilka:
@peti I'm quite sorry but this looks to be an incorrect PR:
* putting valid path (according to Nix) into double quotes results in Nix interpreting `src` as an URL for an archive which ends in something like
    ```
  unpacking sources
  unpacking source archive ./ab
  do not know how to unpack source archive ./ab
    ```
* using tricks like `./. + "/path"` when "path" includes invalid (according to Nix) characters results in an error like
    ```
    invalid character 'Ð' in name 'аб'
    ```
Do you need a PR to revert this one? And sorry once again for not testing it thoroughly.

peti:
I'll revert the patch in https://github.com/NixOS/cabal2nix/pull/404.

qrilka:
Great, thank you.
BTW do you know if there are any documented constraints that Nix sets on paths it could use?
Comments around the line triggering error in `make-derivation.nix` have link to https://github.com/NixOS/nix/pull/1771 but I couldn't find anything explicit about my question in that issue.

PR #405, opened 2019-02-12
Revert "Disable Haddock phase for packages with internal libraries."

shlevy:
Cabal does the right thing since 2.4.0.0 (https://github.com/haskell/cabal/pull/5253)

This reverts commit 7017487bc29bf1d70e38f0ea45878c07db60ab7f.

PR #406, opened 2019-02-18
Revert "Revert "Disable Haddock phase for packages with internal libraries.""

peti:
Reverts NixOS/cabal2nix#405. This change breaks several builds. See https://hydra.nixos.org/build/88843163 for an example.

Cc: @shlevy


akshaymankar:
Is this still the case? I checked one library (`cql-io`) by overriding it with `pkgs.haskell.lib.doHaddock` and it seems to work.

PR #410, opened 2019-03-02
resolveInNixpkgs: stop using GNOME aliases

jtojnar:
They were removed in https://github.com/NixOS/nixpkgs/pull/55735


jtojnar:
Maybe we can get rid of those lines and just rely on the line 143.

ryantm:
There's also gobject-introspection.

ryantm:
Oh, maybe that is going to get handled correctly by using the latest nixpkgs.

jtojnar:
We probably need to run `nix-env -qaP` with `--arg config '{ allowAliases = false; }` to prevent the aliases from being used and tripping OfBorg.

PR #411, opened 2019-03-05
hackage2nix: ignore aliases

jtojnar:
Real solution to https://github.com/NixOS/cabal2nix/pull/410

Depends on https://github.com/peti/distribution-nixpkgs/pull/6

### DIFF
--- a/hackage2nix/Main.hs
+++ b/hackage2nix/Main.hs
@@ -75,7 +75,7 @@ main = do
   CLI {..} <- execParser pinfo
 
   config <- readConfiguration (nixpkgsRepository </>  configFile)
-  nixpkgs <- readNixpkgPackageMap nixpkgsRepository Nothing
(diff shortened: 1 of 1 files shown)


jtojnar:
When I ran `hackage2nix` locally, it fixed the majority of issues. Paging through diff, some new appeared:

* `gnome-keyring` had `libgnome_keyring` changed to `null`
* `webkit2gtk3-javascriptcore` had `webkitgtk` changed to `null` (though some other packages like `webkit-javascriptcore` did not)
* `qr-imager` had `libqrencode` changed to `null`

jtojnar:
Nah, it looks like `pkgs.gnome2.gtk` is still used.

peti:
> When I ran `hackage2nix` locally, it fixed the majority of issues. Paging through diff, some new appeared: [...]

These name lookups are fixed by 50125218b2a2bdafe6bbebf28626b85e73900d51.

peti:
Thank you very much for your help!

PR #416, opened 2019-05-15
Don't hardcode $doc

nh2:
Split outputs can be disabled, in which case $doc is the empty string.

`outputDoc` always points to the right output variable name
(either "out" or "doc").

See also https://github.com/NixOS/nixpkgs/pull/61526.

### DIFF
--- a/src/Distribution/Nixpkgs/Haskell/FromCabal/PostProcess.hs
+++ b/src/Distribution/Nixpkgs/Haskell/FromCabal/PostProcess.hs
@@ -291,8 +291,8 @@ darcsInstallPostInstall = unlines
 xmonadPostInstall :: String
 xmonadPostInstall = unlines
   [ "postInstall = ''"
-  , "  install -D man/xmonad.1 $doc/share/man/man1/xmonad.1"
-  , "  install -D man/xmonad.hs $doc/share/doc/$name/sample-xmonad.hs"
+  , "  install -D man/xmonad.1 ${!outputDoc}/share/man/man1/xmonad.1"
+  , "  install -D man/xmonad.hs ${!outputDoc}/share/doc/$name/sample-xmonad.hs"
   , "'';"
   ]
 


nh2:
Thanks @peti!

peti:
Um, where is that variable `outputDoc` supposed to come from? According to https://hydra.nixos.org/jobset/nixpkgs/haskell-updates#tabs-errors, it appears undefined:

~~~
hydra-eval-jobs returned exit code 1:
error: undefined variable 'outputDoc' at /nix/store/mjadx69m3nwyhiilm46w1ysihnprzcih-source/pkgs/development/haskell-modules/hackage-packages.nix:248972:37
~~~

nh2:
@peti Hmm that is bad, it worked in other places.

It's documented here: https://nixos.org/nixpkgs/manual/#multiple-output-file-type-groups

@matthewbauer are there exceptions to where this works?

src/Distribution/Nixpkgs/Haskell/FromCabal/PostProcess.hs
```diff
@@ -291,8 +291,8 @@ darcsInstallPostInstall = unlines
 xmonadPostInstall :: String
 xmonadPostInstall = unlines
   [ "postInstall = ''"
-  , "  install -D man/xmonad.1 $doc/share/man/man1/xmonad.1"
-  , "  install -D man/xmonad.hs $doc/share/doc/$name/sample-xmonad.hs"
+  , "  install -D man/xmonad.1 ${!outputDoc}/share/man/man1/xmonad.1"
```
matthewbauer: You need two ticks for this to work like:

```
''${!outputDoc}
```

Otherwise Nix thinks it is a variable (as opposed to a bash variable)

nh2: Ah of course, sorry for that @peti. I will PR the fix. I must have copy-pasted it incorrectly from the other nixpkgs editor buffer.

nh2:
Likely fix in #417

PR #417, opened 2019-05-20
xmonad: Fix missing dollar escape quotes.

nh2:
See https://github.com/NixOS/cabal2nix/pull/416#issuecomment-494096785


nh2:
@peti As you can probably tell from the error, I'm making these changes blindly.

Can you teach me how to go all the way from building cabal2nix to running whatever Hydra does?

peti:
@nh2, are you aware of the `update-nixpkgs.sh` script in the `cabal2nix` repo?

nh2:
@peti No, I haven't used any of that stuff yet. Is that the one I should use?

peti:
@nh2, that is the script that's running once per night to update the `haskell-updates` branch. So if you'd like to know how to run `hackage2nix` to re-generate the entire `haskell-packages.nix` file, then that's a pretty good source of information.

PR #435, opened 2019-11-15
Fix pandoc-placetable failure due to dependency on pandoc at setup

unode:
Originally reported at: https://github.com/NixOS/nixpkgs/issues/73392

### DIFF
--- a/src/Distribution/Nixpkgs/Haskell/FromCabal/PostProcess.hs
+++ b/src/Distribution/Nixpkgs/Haskell/FromCabal/PostProcess.hs
@@ -151,6 +151,7 @@ hooks =
   , ("pandoc < 2.6", pandocPre26Overrides)
   , ("pandoc >= 2.6", pandocOverrides)
   , ("pandoc-citeproc", set doCheck False) -- https://github.com/jgm/pandoc-citeproc/issues/369
+  , ("pandoc-placetable", set (libraryDepends . haskell . contains (self "pandoc")) True) -- https://github.com/NixOS/nixpkgs/issues/73392
   , ("purescript", set doCheck False) -- test suite doesn't cope with Nix build env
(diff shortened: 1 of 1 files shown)


cdepillabout:
@unode Thanks for sending a fix for this.

However, it appears I was incorrect about putting `pandoc` in library depends.  It should actually be in the executable depends section.

Here's what the .nix file looks like when using `cabal2nix` to generate it for `pandoc-placetable` with this PR:

```
{ mkDerivation, aeson, base, bytestring, explicit-exception
, http-conduit, pandoc, pandoc-types, spreadsheet, stdenv, text
, utf8-string
}:
mkDerivation {
  pname = "pandoc-placetable";
  version = "0.5.1";
  sha256 = "22844b30fa5be4190e816de920d5bbbdccde18e4022c7927a97c4b83b0aed87d";
  configureFlags = [ "-finlineMarkdown" ];
  isLibrary = false;
  isExecutable = true;
  enableSeparateDataOutput = true;
  libraryHaskellDepends = [ pandoc ];
  executableHaskellDepends = [
    aeson base bytestring explicit-exception http-conduit pandoc-types
    spreadsheet text utf8-string
  ];
  homepage = "https://github.com/mb21/pandoc-placetable";
  description = "Pandoc filter to include CSV files";
  license = "GPL";
}
```

So it looks like it is working as expected, but `pandoc` needs to be in the `executableHaskellDepends` section.

peti:
I believe that the underlying issue causing the missing dependency was fixed by 650d5300c6bb5632bbab6fcb2d939fb476e469c5?

cdepillabout:
@peti I believe you're correct.

https://github.com/NixOS/nixpkgs/pull/74050 removes `pandoc-placetable` from the list of broken packages in nixpkgs.

So I think this PR can be closed.

PR #444, opened 2020-03-14
secp256k1-haskell: fix pkgconfig-depends

jtobin:
This library is currently broken in nixpkgs; overriding 'libsecp256k1' with 'secp256k1' in pkgconfig-depends unbreaks it.

I've tested that this works as expected by building `cabal2nix`, calling `cabal2nix cabal://secp256k1-haskell-0.1.8`, and running `nix-build` on the (appropriately-called) result.

See NixOS/nixpkgs#82562.


cdepillabout:
@jtobin Thanks for following up here.

I _think_ the correct place to make this change is in this list:

https://github.com/NixOS/cabal2nix/blob/722a3771a70d8f9221aa77c30fb9aea7fb1df6c6/src/Distribution/Nixpkgs/Haskell/FromCabal/Name.hs#L16-L31

This is the list that maps pkg-config dependencies to their respective package name in nixpkgs.

Adding the correct mapping for `libsecp256k1` to this list will make it so that any Haskell package that uses `libsecp256k1` gets the correct system package in nixpkgs.

jtobin:
Nice catch, thanks.  Done, and confirmed as working via the same test mentioned in OP.

jtobin:
(Oh, Dennis!  Hi!  We chatted in late 2018.  Hope all's well! :smile:)

PR #452, opened 2020-04-21
[ cleanup ] don't build Agda builtins again

laMudri:
Since https://github.com/agda/agda/commit/378020321a223d7057cd10b59b995ebac91c7a8b, Agda's Setup.hs has contained the instructions for building the builtin modules. Thus, we do not need to do it again.

I haven't tested this yet, because building Agda requires slightly more RAM than my machine has. I can test it overnight if required, though.

/cc @abbradar


peti:
Well, I'd be happy to merge the PR, but would be nice if you could make sure that it actually works (e.g. please *test it*).

peti:
I checked it, and Agda compiles fine with your changes applied. Thank you very much!

Issue #431, opened 2019-10-13
hackage2nix reproducibility


roberth:
I've tried to run hackage2nix in order to test a change I made to the yaml file in Nixpkgs. I didn't make it through. I'll summarize my experience.

1. I used `git add worktree` and `git checkout --detach` instead of `git clone nixpkgs` to speed it up. Didn't work because `update-nixpkgs.sh` pulls in the nixpkgs worktree.
2. I had to add `cabal`, `ghc`, `zlib` and `openssl` to my shell before invoking `update-nixpkgs.sh`
3. Still, cabal failed with
    ```
    $ ./update-nixpkgs.sh 
    <command line>: can't load .so/.DLL for: libz.so (libz.so: cannot open shared object file: No such file or directory)
    cabal: Failed to build cabal2nix-2.15.0 (which is required by exe:hackage2nix
    from cabal2nix-2.15.0).
    ```

# Proposed solution

I think the script can be nixified. I suggest adding the following to Nixpkgs:

 - pin files for
   - cabal2nix repo revision
   - all-cabal-hashes repo revision
 - a port of `update-nixpkgs.sh`, limited to generation
   - using the pins
   - no assumption about the caller's environment (nix-shell shebang or similar)
   - no git logic, just generation
 - a script to update the pins and call the codegen script
 - assuming the code generation is now pure, a derivation that tests whether the generated code is up to date, to prevent hand-editing and ensure reproducibility

This will let anyone regenerate the package set in a reproducible way. This also makes #219 trivial. Just mention the pin files in a comment.

Did I miss something that makes the above not work?

peti:
https://www.youtube.com/watch?v=qX0mgtSm360 has very detailed instructions how to run `hackage2nix`. Maybe that is helpful? Once you have actually built the executable correctly, you'll find that it's actually pretty easy to run.

roberth:
That's really cool that you've put it in a video. Video is a great format for explaining things.

When I tried running the update script, I followed the Setup and Run steps in the [hackage2nix README](https://github.com/NixOS/cabal2nix/tree/c6f572cc24b9b145b8cc0cf53dbcd489b81027fc/hackage2nix). My goal was to test a configuration change, not hacking on hackage2nix or anything like that. Even without my "proposed solution" I don't think that it should take more than a little shell script invocation.

If my intention was to do more work on hackage2nix etc, I'd probably have found the video.

peti:
[...]

PR #453, opened 2020-04-26
Improve READMEs, reproducibility, and make contributing easier

nh2:
This is my first shot at solving #431 and making it easier for more people to contribute to `stack2nix`, `stackage2nix`, and the nixpkgs Haskell package set upgrades.

I still have to make a less customised version of `./update-nixpkgs.sh` (e.g. one that doesn't push stuff), but that is for a later PR.

### DIFF
--- a/src/Distribution/Nixpkgs/Haskell/FromCabal.hs
+++ b/src/Distribution/Nixpkgs/Haskell/FromCabal.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards #-}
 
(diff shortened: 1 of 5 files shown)


src/Distribution/Nixpkgs/Haskell/FromCabal.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
```
peti, line 1: I really don't want to support multiple versions of Cabal, especially if that support requires CPP. I am sorry, but `cabal2nix` is inherently tied to Nixpkgs and it's very easy to build this tool within Nixpkgs. Adding complexity to the build to enable additional methods to build this tool is just not a priority for me.

nh2: @peti I understand that was your original motivation, but I firmly believe that being compatible with the latest LTS will get you many more contributors (such as Robert from #431, me, and so on), and that the time you will save by having people help out will dwarf the additional effort (e.g. between Cabal 3.0 and 3.2, it is only 2 lines).

Also, the differences would likely be quite small, as the _latest_ LTS is enough (we don't need to keep compat with older ones).

I propose to roll with it for now, we can still throw it away again if we do find that the maintenance burden is large.

peti: OK.

PR #456, opened 2020-05-25
Use update scripts from nixpkgs

hyperfekt:
Changes the automated update script to call the one from nixpkgs.
Also adds one that uses the local checkout of cabal2nix but does not commit.
A less invasive alternative is proposed in #457


peti:
No, I am sorry, but there is no way I'm adding scripts to this repository that call scripts in another repository to do the work that used to be done by these scripts themselves. This is way too much indirection and complexity for my taste and I don't want to work like this.

PR #457, opened 2020-05-25
extend nixpkgs update scripts to pin used versions

hyperfekt:
Makes sure the update scripts leave behind the necessary pins for the regenerate-hackage-packages maintainer script in nixpkgs to work.

### DIFF
--- a/update-hackage-snapshot.sh
+++ b/update-hackage-snapshot.sh
@@ -5,12 +5,7 @@ set -eu -o pipefail
 rev=$( git ls-remote --heads git://github.com/commercialhaskell/all-cabal-hashes.git hackage \
      | cut -f1
      )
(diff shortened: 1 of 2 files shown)


peti:
I am sorry, @hyperfekt, but I don't have any use for these pinned version files. I am in favor of implementing an option to re-produce an older build with the generator scripts, but I would like that to be implemented by parsing the commit messages, which is really simple because all the information is baked into a simple URL in there.

hyperfekt:
It's not as much a matter of difficulty as one of principle. I strongly believe Git commit messages should not be a functional part of nixpkgs.

hyperfekt:
If the problem is the time taken by the nix-prefetch-github command, I'd be happy to modify the PR to compute the hash from a checkout instead.

PR #462, opened 2020-06-14
Various improvements to ./regenerate-nixpkgs.sh

maralorn:
* Ability to checkout cabal2nix tags
* Use nix-shell
* Clone all-cabal-hashes if necessary
* More helpful output

I couldn't figure out how to install cabal2nix.env without cabal2nix itself via the shebang, but I believe that this still provides the dependencies to cabal.


peti:
Here is one more suggestion:

    $ cabal --with-compiler="$(cat $(nix-build --no-out-link "<nixpkgs>" -A haskellPackages.cabal2nix.env))/bin/ghc" run hackage2nix -- dflags auth oa stuho

PR #477, opened 2021-01-11
Use lib instead of stdenv.lib

rnhmjoj:
On Nixpkgs, the work[1] for deprecating stdenv.lib has started.
This patch updates the generator, existing documentation and tests to
use the top-level lib attribute to access the standard library, instead
of stdenv.lib.

Note: to update the test cases I did the following:

    $ cabal2nix . > default.nix
    $ sed -i '35i doCheck = false;' default.nix
    $ nix-build
    $ find test/golden-test-cases/ -name '*.cabal' | \
      sed s/.cabal//g | xargs -n1 -I fname \
      sh -c 'result/bin/cabal2nix file://fname.cabal > fname.nix.golden'
    $ find test/golden-test-cases/ -name '*.nix.golden' | \
      xargs sed -e 's#sha256 = "\w\\+#sha256 = "deadbeef#g' -i

[1]: https://github.com/NixOS/nixpkgs/issues/108938


peti:
@rnhmjoj , you can re-generate the golden files for the regression test suite by running `cabal run regression-test -- --accept`. It's a wee bit easier than constructing a crazy `sed` command-line every you wanna change cabal2nix's output. :-)

src/Distribution/Nixpkgs/Haskell/Derivation.hs
```diff
@@ -110,7 +110,7 @@ instance Package Derivation where
 instance NFData Derivation
 
 instance Pretty Derivation where
-  pPrint drv@MkDerivation {..} = funargs (map text ("mkDerivation" : toAscList inputs)) $$ vcat
+  pPrint drv@MkDerivation {..} = funargs (map text ("lib" : "mkDerivation" : toAscList inputs)) $$ vcat
```
peti, line 113: The `lib` argument does not belong here.

`mkDerivation` is a very special argument, because it's the function that actually constructs a derivation from the given attribute set. We customize the Haskell package by overriding that function to mess with the attribute set in some way, i.e. enabling by enabling shared library builds or disabling test suites, etc. Therefore, it's really important that all Haskell packages take that function as an argument so that we can override it.

`lib`, on the other hand, is not something that we'll ever want to override for a single Haskell derivation, therefore it's not necessary to pass it to every single build individually. Instead, we should pass `lib` (instead of `stdenv`) at the top of `pkgs/development/haskell-modules/hackage-packages.nix` globally for *all* Haskell derivations.

peti (changes requested):

rnhmjoj:
> t's a wee bit easier than constructing a crazy sed command-line every you wanna change cabal2nix's output. :-)

Thank you, I expected you had a better way to do it.

src/Distribution/Nixpkgs/Haskell/Derivation.hs
rnhmjoj: > mkDerivation is a very special argument

Yeah, I knew this wasn't the right place: I would have replaced `stdenv` with `lib` but I don't see it appearing in this repository at all. How does it work?

> Instead, we should pass lib (instead of stdenv) at the top of pkgs/development/haskell-modules/hackage-packages.nix globally for all Haskell derivations.

What about standalone expressions?

peti: > I would have replaced stdenv with lib but I don't see it appearing in this repository at all. How does it work?

`stdenv` is an argument to the entire package set in `hackage-packages.nix`. The code is generated here: https://github.com/NixOS/cabal2nix/blob/master/hackage2nix/Main.hs#L178

> What about standalone expressions?

Standalone expressions get `stdenv` added to their list of expected arguments by cabal2nix. The code is at: https://github.com/NixOS/cabal2nix/blob/master/src/Cabal2nix.hs#L241

rnhmjoj: Thank you, that helps me at lot! I'm on it.

rnhmjoj:
I fixed the problems above but there's still one issue: the rendering of the meta section is handled in `Distribution.Nixpkgs`, so that will have to change as well. I'll open a PR soon.

rnhmjoj:
See https://github.com/peti/distribution-nixpkgs/pull/7

PR #480, opened 2021-01-29
Document you can use a 'src' attribute in hackage-packages.nix

raboof:


peti (changes requested):
It's typically a bad idea to edit `hackage-packages.nix` directly and I would rather not suggest that people do that in the official documentation. The proper way to change the `src` attribute of an existing derivation is to create an override with `overrideSrc` function. You can look at the case of `git-annex` for an example: https://github.com/NixOS/nixpkgs/blob/99eeda9fdbd36022694f413f9d79fd626b95e849/pkgs/development/haskell-modules/configuration-common.nix#L62-L68

raboof:
> The proper way to change the `src` attribute of an existing derivation is to create an override with `overrideSrc` function.

Thanks, updated!

PR #483, opened 2021-02-22
Update hackage2nix instructions to reflect cabal v2 usage

sternenseemann:
Seems like this README was not updated when stack was removed from this
repository and the old style cabal workflow was dropped in favor of
cabal v2-*. Now we need to build cabal2nix using cabal v2-build in order
for the cabal v2-run in update-nixpkgs.sh to work.


Jake-Gillberg:
I also needed to get `cabal` on my path with a `nix-shell -p cabal-install`

PR #484, opened 2021-03-09
feat(FromCabal.License): don't generate unclear attributes for gpl licenses

sternenseemann:
In nixpkgs, the licenses.{gpl3,gpl2,lgpl21,…} attributes are slowly
being deprecated in favor of the licenses.{gpl3Only, gpl2Only, …}
attributes which avoids confusion with the licenses.{gpl3Plus, gpl2Plus, …}
attributes. We should probably do the same as we are already doing with
the SPDX license identifiers.


sternenseemann:
Updated test cases as well now :)

PR #490, opened 2021-04-29
FromCabal.Name: use linux-pam for pam

nkpvk:
`pam` is an alias for `linux-pam` in Nixpkgs. This causes a build failure for `haskellPackages.pam`.

Reference https://github.com/NixOS/nixpkgs/pull/121049


src/Distribution/Nixpkgs/Haskell/FromCabal/Name.hs
```diff
@@ -125,6 +125,7 @@ libNixName "nix-store"                          = return "nix"
 libNixName "notify"                             = return "libnotify"
 libNixName "odbc"                               = return "unixODBC"
 libNixName "openblas"                           = return "openblasCompat"
+libNixName "pam"                                = return "linux-pam"
```
sternenseemann, line 128: I think it'd be better to use the `pam` attribute since that also works on darwin (on linux it's an alias for linux-pam).

However, I thought that `cabal2nix` just uses the name specified in the cabal file as is if it exists. I'll check tomorrow if nobody else can answer this for now.

nkpvk: So, the problem is `pam` being set to `null`:

```nix
"pam" = callPackage
  ({ mkDerivation, base, c2hs, pam }:
   mkDerivation {
     pname = "pam";
     ...
     librarySystemDepends = [ pam ];
     ...
   }) {pam = null;};
```

Adding something like this:

```hs
hooks :: [(Dependency, Derivation -> Derivation)]
hooks =
  ...
  , ("pam", over (libraryDepends . system) (replace "pam = null" (pkg "pam")))
  ...
```

to `FromCabal/PostProcess.hs`, produces the following:

```nix
"pam" = callPackage
  ({ mkDerivation, base, c2hs, pam }:
   mkDerivation {
     pname = "pam";
     ...
     librarySystemDepends = [ pam ];
     ...
   }) {inherit (pkgs) pam;};
```

and makes the build succeed. This should now also work on darwin, I hope.

sternenseemann: Okay, the more I look into this, the more I am convinced that this _should_ already work without any postprocessing. It seems like the `nixpkgsResolver` in `hackage2nix` does the wrong thing currently (return `python38Packages.pam`) which is rejected by `goodScopes` in `FromCabal.hs`.

sternenseemann: I think <https://hackage.haskell.org/package/distribution-nixpkgs-1.5.0/docs/Distribution-Nixpkgs-PackageMap.html#v:resolve> is to blame.

sternenseemann: The issue is that `nix-env -qaP --json` is used to get a list of all atribute paths from nixpkgs. This doesn't list all attributes unfortunately, only one attribute per unique derivation, so only linux-pam is printed and not pam.

I'll look into a way of fixing this!

sternenseemann: See <https://github.com/peti/distribution-nixpkgs/issues/9>. I think in the mean time its easiest to add an

```nix
  pam = super.pam.override { inherit (pkgs) pam; };
```

to `configuration-common.nix` as a workaround.

sternenseemann:
@nkpvk See #492.

PR #491, opened 2021-05-03
doc: Explain how to use GHC with LLVM

Mathnerd314:
It's complicated and requires various incantations, and isn't documented anywhere currently that I could find.

I'm not sure what the difference between the user's guide and FAQ is, but I went with the FAQ.

### DIFF
--- a/doc/frequently-asked-questions.rst
+++ b/doc/frequently-asked-questions.rst
@@ -449,6 +449,51 @@ following to get the ``scientific`` package build with
 
    nix-build -A haskell.packages.integer-simple.ghc802.scientific
 
+How to get an environment with GHC and LLVM
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By default LLVM is not included with GHC unless it is required for native code generation:
+
+.. code:: shell
+
+   $ nix-shell -p 'haskellPackages.ghcWithPackages (ps: with ps; [mtl])'
+
+   [nix-shell:~]$ ghc -fllvm Main.hs
+   <no location info>: error:
+       Warning: Couldn't figure out LLVM version!
+                Make sure you have installed LLVM 9
+   ghc: could not execute: opt
+   
+   [nix-shell:~]$ llvm-ar
+   llvm-ar: command not found
+  
+To include LLVM in GHC you have to override the expression:
+
+.. code:: shell
+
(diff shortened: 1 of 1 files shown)


doc/frequently-asked-questions.rst
```diff
@@ -449,6 +449,51 @@ following to get the ``scientific`` package build with
 
    nix-build -A haskell.packages.integer-simple.ghc802.scientific
 
+How to get an environment with GHC and LLVM
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By default on x86_64 LLVM is not included with GHC:
```
sternenseemann, line 455:
```suggestion
By default LLVM is not included with GHC unless it is required for native code generation:
```

(Although this will only be true once we merge the new update)

doc/frequently-asked-questions.rst
```diff
@@ -449,6 +449,51 @@ following to get the ``scientific`` package build with
...
+.. code:: shell
+
+   [nix-shell:~]$ llvm-ar
+   llvm-ar: command not found
+
+The solution is to add it as another package to the environment:
+
+.. code:: shell
+
+   $ nix-shell -p '(haskellPackages.ghcWithPackages (ps: with ps; [mtl])).override { withLLVM = true; }' 'haskellPackages.llvmPackages.llvm'
```
sternenseemann, line 491:
```suggestion
   $ nix-shell -p '(haskellPackages.ghcWithPackages (ps: with ps; [mtl])).override { withLLVM = true; }' 'llvmPackages.llvm'
```

Mathnerd314: Well, the idea here is that you want to use the same LLVM version as GHC is using. So this change isn't correct, this would give LLVM 7 instead of LLVM 9.

sternenseemann: Oh, nevermind me, I didn't know about this reexposure of llvm.

Mathnerd314: I mean, you could use `llvm_9`, but looking up the version of LLVM for the version of GHC you're using is a pain.

sternenseemann: Yeah, no, I agree this is much better how you had it. I just wasn't aware of the re-exposure and thought it was a typing mistake.

PR #494, opened 2021-05-05
hackage2nix: Accept multiple config files

maralorn:
This commit is preparing the switch to multiple config files in nixpkgs.

I will soon create a PR to nixpkgs demonstrating the usage of multiple config files.


maralorn:
I haven‘t figured out how to set a default value for a some1 value. But it doesn‘t matter much. We only use this script with specific supplied config and I plan to remove that default value anyways.

sternenseemann (approved):

PR #495, opened 2021-05-22
doc: mention using Lorri with haskell-language-server

Mathnerd314:
See https://github.com/haskell/haskell-language-server/issues/493. This seems to be the recommended way to setup the language server for multiple GHCs / projects.

### DIFF
--- a/doc/nixpkgs-users-guide.rst
+++ b/doc/nixpkgs-users-guide.rst
@@ -473,6 +473,14 @@ specific nix-shell You can either use an override as given above or
 simply install ``pkgs.haskellPackages.haskell-language-server`` instead
 of the top-level attribute ``pkgs.haskell-language-server``.
 
+How to use haskell-language-server in an editor with a GHC from nix-shell
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This is solved fairly well with direnv:
+
+* Use `lorri <https://github.com/target/lorri>`__ to get information from the nix-shell to direnv
+* Use a direnv support plugin in your editor (emacs has one, vscode has one) to get the right environment for the server launch.
+
 How to build a Haskell project using Stack
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/doc/nixpkgs-users-guide.rst
+++ b/doc/nixpkgs-users-guide.rst
@@ -473,13 +473,29 @@ specific nix-shell You can either use an override as given above or
 simply install ``pkgs.haskellPackages.haskell-language-server`` instead
 of the top-level attribute ``pkgs.haskell-language-server``.
 
-How to use haskell-language-server in an editor with a GHC from nix-shell
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+How to make haskell-language-server find a GHC from nix-shell
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-This is solved fairly well with direnv:
+If you use nix-shell for your development environments then ``haskell-language-server``
+will not find an installed GHC or will find a GHC with an installed package set
+different from what your project uses.
 
-* Use `lorri <https://github.com/target/lorri>`__ to get information from the nix-shell to direnv
-* Use a direnv support plugin in your editor (emacs has one, vscode has one) to get the right environment for the server launch.
+The simplest solution to this problem is to launch your editor from within the nix-shell
+environment:
+
+.. code:: shell
+
+  $ nix-shell
+  [nix-shell] $ code .
+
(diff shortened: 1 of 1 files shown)


maralorn:
First: Thanks for improving the docs!

---

I think we need to be very precise about what problem we want to solve here:

1) If you use nixos imo the recommended way would be to install `pkgs.haskell-language-server` system wide. Then thanks to haskell-language-server-wrapper, every project can pick it’s correct ghc. No nix-shell needed in most situations.

2) If you want to install haskell-language-server from your nix-shell, then the most important point is that you need to run your editor from within the nix-shell. That is everything needed to hit the ground running.

3) We can recommend using either lorri _or_ nix-direnv for automatically entering the nix-shell, when you enter the directory.

4) We can recommend using an editor plugin for auto loading the direnv.

---

I don‘t necessarily expect you to explain all of this in this PR. But before we merge I think the documentation should to be somehow clear on two things:

* `2)` is enough for a working setup. I don‘t want that people look at this and either shy away from the complexity or choose a solution that has a higher complexity than is good for their current skill level.
* I would prefer suggesting https://github.com/nix-community/nix-direnv, alongside lorri because it is conceptually simpler and more lightweight (to setup).

doc/nixpkgs-users-guide.rst
```diff
@@ -473,6 +473,14 @@ specific nix-shell You can either use an override as given above or
 simply install ``pkgs.haskellPackages.haskell-language-server`` instead
 of the top-level attribute ``pkgs.haskell-language-server``.
 
+How to use haskell-language-server in an editor with a GHC from nix-shell
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This is solved fairly well with direnv:
+
+* Use `lorri <https://github.com/target/lorri>`__ to get information from the nix-shell to direnv
```
sternenseemann, line 481: Is maintained in the nix-community organization nowadays.

```suggestion
* Use `lorri <https://github.com/nix-community/lorri>`__ to get information from the nix-shell to direnv
```

Mathnerd314:
> I think we need to be very precise about what problem we want to solve here.

Well, the situation I'm in is that every project runs in its own nix-shell (simple ghcWithPackages environments) and I don't have any GHC installed system-wide.  So just installing haskell-language-server (method 1) doesn't work because it doesn't find any GHC. Even if I added a GHC to the system path I don't think it would work because it wouldn't see the packages installed for each project.

Method 2 does work to get a GHC, and AFAICT it even works with buildFHSUserEnv, so that should probably be the recommended method.

lorri, nix-direnv, etc. can all be grouped as "direnv methods" so I think the solution is to link to https://github.com/direnv/direnv/wiki/Nix and let people figure it out themselves (there is a features table there).

There is also an extension https://github.com/arrterian/nix-env-selector/ based on `nix-shell --run`, I guess it also deserves a mention. Are there any similar extensions for other editors?

maralorn:
> Well, the situation I'm in is that every project runs in its own nix-shell (simple ghcWithPackages environments) and I don't have any GHC installed system-wide. So just installing haskell-language-server (method 1) doesn't work because it doesn't find any GHC. Even if I added a GHC to the system path I don't think it would work because it wouldn't see the packages installed for each project.

Fair point. I have a global ghc, a global hls and packages either global too or via cabal. So for me it works. But I see that people will often want the nix-shell solution.

> Method 2 does work to get a GHC, and AFAICT it even works with buildFHSUserEnv, so that should probably be the recommended method. Although when I tried it just now VSCode didn't load the theme right.

Having a nix-shell for your project is obviously very widespread and very versatile. (As I side note: we can probably try to iterate on the shell.nix examples in this guide. There are often blog posts making recommendations on this and they vary a lot in quality and creativity. Why not make our documentation on this so good that we can point people their and say: That’s what you want.)

> lorri, nix-direnv, etc. can all be grouped as "direnv methods" so I think the solution is to link to [direnv/direnv/wiki/Nix](https://github.com/direnv/direnv/wiki/Nix) and let people figure it out themselves (there is a features table there).

Yep, very good!

> There is also an extension [arrterian/nix-env-selector](https://github.com/arrterian/nix-env-selector/) based on `nix-shell --run`, I guess it also deserves a mention. Are there any similar extensions for other editors?

Interesting. I don‘t know.

---

I was maybe a bit general yesterday so I made a rather meandering comment. (Which we seem mainly to agree on.) Let me make my change request more precise: I see a mismatch between your heading and the content. People could read it as "direnv" is the only way to load your nix-shell into an editor. You should either put an "automatically" in the title or an "You can just run nix-shell and start the editor from within it. If you want that to happen automatically, then …".

Of course the whole thing is a bit not hls specific so I wonder if it’s a good idea to have a whole subsection for it. Maybe just a sentence? Maybe we should actually elaborate on this elsewhere in the document where we describe how to setup a dev shell? But that’s subject to bike shedding so I’ll let you decide.

maralorn:
Nice! Thank you!

doc/nixpkgs-users-guide.rst
Mathnerd314: I updated https://github.com/direnv/direnv/wiki/Nix. But they really need to update the README in the target repo...

Profpatsch: Point taken!

PR #508, opened 2021-06-30
Don't use hpack when cabal file is in the directory

### LINKED ISSUE #507: cabal2nix can fail if the `hpack` version is newer
I'm testing a build out and depending on a repository. I invoked `cabal2nix` like so:

```
cabal2nix https://github.com/parsonsmatt/persistent-discover --revision 327106b1a74d68def6f765e4df435d3f567163bc > .nix/deps/persistent-discover.nix
```

The output when trying to enter `nix-shell` is this:

```
$ nix-shell
warning: dumping very large path (> 256 MiB); this may run out of memory
these derivations will be built:
  /nix/store/j5ym74w47qa1q7j6fs47dfn77bqrgly1-persistent-discover-0.1.0.3.drv
  /nix/store/fw8ps5sal8f6mwgim7389fb6qv4nwa0q-hoogle-local-0.1.drv
  /nix/store/klij767lxkll2w91hg6i1a2sw4q12n6g-ghc-8.10.3-with-packages.drv
building '/nix/store/j5ym74w47qa1q7j6fs47dfn77bqrgly1-persistent-discover-0.1.0.3.drv'...
setupCompilerEnvironmentPhase
Build with /nix/store/bsr0vynkkbz2rmhzf6x64yahigyydj7c-ghc-8.10.3.
unpacking sources
unpacking source archive /nix/store/wfy2cwydn33ql1za6f30a08282s8ma5p-persistent-discover-327106b
source root is persistent-discover-327106b
patching sources
persistent-discover.cabal was generated with a newer version of hpack, please upgrade and try again.
builder for '/nix/store/j5ym74w47qa1q7j6fs47dfn77bqrgly1-persistent-discover-0.1.0.3.drv' failed with exit code 1
cannot build derivation '/nix/store/klij767lxkll2w91hg6i1a2sw4q12n6g-ghc-8.10.3-with-packages.drv': 1 dependencies couldn't be built
error: build of '/nix/store/klij767lxkll2w91hg6i1a2sw4q12n6g-ghc-8.10.3-with-packages.drv' failed
```

The error line is this:

> `persistent-discover.cabal was generated with a newer version of hpack, please upgrade and try again.`

I've worked around this by adding `--force` to our `hpack` invocation at work, so `hpack --force .` will allow it to override a thing. `cabal2nix` should probably do this too, in the event that a Git repo has a newer `hpack` than the one built in to the environment.

parsonsmatt:
Closes #507 

This PR makes `hpack` force generation of a `cabal` file. `hpack` will bail out early if the version of `hpack` used to generate an existing `cabal` file is newer than itself. This results in a package that can't be built.

### DIFF
--- a/src/Distribution/Nixpkgs/Haskell/PackageSourceSpec.hs
+++ b/src/Distribution/Nixpkgs/Haskell/PackageSourceSpec.hs
@@ -197,7 +197,7 @@ cabalFromDirectory :: HpackUse -- ^ the way hpack should be used
 cabalFromDirectory ForceHpack dir = hpackDirectory dir
 cabalFromDirectory NoHpack dir = onlyCabalFromDirectory dir "*** No .cabal file was found. Exiting."
 cabalFromDirectory PackageYamlHpack dir = do
-  useHpack <- liftIO $ doesFileExist (dir </> "package.yaml")
+  useHpack <- liftIO $ shouldUseHpack dir
   if useHpack
     then do
       liftIO $ hPutStrLn stderr "*** found package.yaml. Using hpack..."
@@ -206,11 +206,29 @@ cabalFromDirectory PackageYamlHpack dir = do
 
 onlyCabalFromDirectory :: FilePath -> String -> MaybeT IO (Bool, Cabal.GenericPackageDescription)
 onlyCabalFromDirectory dir errMsg = do
-  cabals <- liftIO $ getDirectoryContents dir >>= filterM doesFileExist . map (dir </>) . filter (".cabal" `isSuffixOf`)
+  cabals <- liftIO $ findCabalFiles dir
   case cabals of
     [] -> liftIO $ fail errMsg
     [cabalFile] -> (,) False <$> cabalFromFile True cabalFile
     _ -> liftIO $ fail ("*** found more than one cabal file (" ++ show cabals ++ "). Exiting.")
+
+-- | Returns a list of files ending with the @.cabal@ suffix.
+findCabalFiles :: FilePath -> IO [FilePath]
+findCabalFiles dir = do
+  contents <- getDirectoryContents dir
+  filterM doesFileExist . map (dir </>) . filter (".cabal" `isSuffixOf`) $ contents
+
+-- | This function returns 'True' if a @package.yaml@ is present and there
+-- are no @.cabal@ files in the directory.
+shouldUseHpack :: FilePath -> IO Bool
+shouldUseHpack dir = do
+  hpackExists <- doesFileExist (dir </> "package.yaml")
+  if hpackExists
+    then do
+      cabalFiles <- findCabalFiles dir
+      pure $ null cabalFiles
+    else
+      pure False
 
 handleIO :: (Exception.IOException -> IO a) -> IO a -> IO a
 handleIO = Exception.handle


cdepillabout (approved):

sternenseemann:
I'm not in favor of this and also suggest either hpack should be disabled by the user in such cases or the updated cabal file checked in. If the `--force` is desired, this can easily be achieved via an override. We also could add an argument to `haskellPackages.mkDerivation` like `hpackFlags` to make this even easier.

cdepillabout:
> I'm not in favor of this and also suggest either hpack should be disabled by the user in such cases or the updated cabal file checked in. 

What's you're reasoning for this?  It seems to me like this is in general an improvement.

sternenseemann:
If a newer hpack version was used for generating the cabal file, this implies that newer features not supported by our hpack version may have been used. Therefore this is a serious warning in my opinion, indicating that we may generate an incorrect cabal file.

cdepillabout:
> this implies that newer features not supported by our hpack version may have been used.

My guess is that this is not a big problem in practice.

I think we have these three different cases to deal with:

1. `package.yaml` was generated by a newer version of `hpack`, but Nix uses an older version of `hpack`.  The `package.yaml` file doesn't use any features from the newer version of `hpack`, so generating the `.cabal` file works with the older version of `hpack`.

    This is specifically the case where `--force` would help.  I'm guessing this case would happen _way_ more often than the other cases.

1. `package.yaml` was generated by a newer version of `hpack`, but Nix uses an older version of `hpack`.  The `package.yaml` file uses a new feature from the newer version of `hpack`.

    In this case, regardless of whether we used the `--force` flag or not, compilation still fails (since we can't generate a `.cabal` file).  We are no worse here because of `--force`.

1.  `package.yaml` was generated by a newer version of `hpack`, but Nix uses an older version of `hpack`.  The `package.yaml` file uses a new feature from the newer version of `hpack`.  The older version of `hpack` is still able to compile the `package.yaml` file though, but this produces a `.cabal` file that isn't able to be compiled correctly.

    From an end-user perspective, they see their package still not compiling regardless of whether or not `--force` is used.  We are no worse here because of `--force`.

1. `package.yaml` was generated by a newer version of `hpack`, but Nix uses an older version of `hpack`.  The `package.yaml` file uses a new feature from the newer version of `hpack`.  The older version of `hpack` is still able to compile the file though.  This produces a `.cabal` file that _is_ able to be compiled correctly, but this causes weird errors at runtime.

    I think this is the only "dangerous" outcome of using `--force`.  I imagine this is pretty unlikely to happen, though.  I'd be pretty surprised if anyone could come up with a realistic example of a `package.yaml` file where this occurred.

--------------------------------------

I think case 1 is by far what is mostly likely to happen, so using `--force` by default seems like it would just make a smoother workflow for most users.  My guess is that the likelihood this causes problems is pretty low.

maralorn:
> I'm not in favor of this and also suggest either hpack should be disabled by the user in such cases or the updated cabal file checked in.

I think forcing the user to check in a downgraded hpack file sounds like a weird solution.

I think what we actually want is this.

1. Expose an easy way to allow `hpack --force`.
2. Fail when the `--force` is needed.
3. Show an easy to understand error message that explains what is going on and how to fix your derivation.

But well, that’s not easy to achieve. I am not sure what bullet to bite here. But I actually think the probability of having a user loose 2h in figuring this out or just giving up is higher than the probability of us quietly compiling in a wrong way.

parsonsmatt:
Another option is to just *not* run `hpack` if there's a `.cabal` file already. It's considered [good practice to store generated cabal files](https://www.fpcomplete.com/blog/storing-generated-cabal-files/). I ran into this problem because I checked the cabal file in to the repository.

Looks like [the default option](https://github.com/NixOS/cabal2nix/blob/e18310f2eae5d2d7b73512b0b75396224fbf76ca/src/Cabal2nix.hs#L80-L84) is to use `hpack` if [there is a `package.yaml` file](https://github.com/NixOS/cabal2nix/blob/3d9f69c17bbb82546d747906a83dbf0c2ddfa0d5/src/Distribution/Nixpkgs/Haskell/PackageSourceSpec.hs#L199-L205). Maybe the logic should be to only use `hpack` *if* there is *only* a `package.yaml` file, and trust that a checked-in `.cabal` file is intended to be the source of truth.

sternenseemann:
> Looks like the default option is to use hpack if there is a package.yaml file. Maybe the logic should be to only use hpack if there is only a package.yaml file, and trust that a checked-in .cabal file is intended to be the source of truth.

Yeah, that sounds like the most sensible way forward. The logic is a bit more complicated though since the cabal file's name changes with the package name (can there even be multiple ones in a single directory?).

parsonsmatt:
`cabal` errors out if there are multiple `.cabal` files in a directory.

```
λ ~/Projects/persistent-discover/ cp persistent-discover.cabal persistent-lmao.cabal
λ ~/Projects/persistent-discover/ cabal build 
When using configuration(s) from /home/matt/Projects/persistent-discover/cabal.project, the following errors occurred:
The package directory '.' contains multiple .cabal files (which is not currently supported).
```

I've updated the PR to not run `hpack` if there is a `cabal` file present in the directory.

cdepillabout:
I've tested that this works as advertised.

If no `.cabal` file exists, then `package.yaml` is used.  If a `.cabal` file exists, then `package.yaml` is not used regardless of whether or not it exists.

@maralorn @sternenseemann any objection to merging this?

maralorn (approved):
LGTM. Thank you!

Since this changes user facing behavior I think this should probably have a changelog entry, because it could theoretically break projects with out-dated checked-in cabal files. But apparently this project has no changelog. So whatever, I guess?^^

sternenseemann:
My plan was to release a bunch of non-breaking changes and fixes as a minor release (<https://github.com/NixOS/cabal2nix/milestone/1>), we'd be able to backport. If we still want to do that, we should push this PR into <https://github.com/NixOS/cabal2nix/milestone/2> and hold off on merging for now — unless we want to maintain two branches in this repository

I think the minor release would be great because it would be backport-able and includes a fix to the annoying `callCabal2nix` error message problem. However, I currently can't do anything to accelerate this process since it is blocked on <https://github.com/NixOS/nixpkgs/issues/125980>.

sternenseemann:
Rebased on master and squashed the commits. Would be fine with merging later!

pwm:
Just my 2c: while I think this is ultimately a good change I think it should have been marked as breaking in the changelog. Previously the workings of `cabal2nix` was independent from the existence of .cabal files if `package.yaml` was present whilst now it depends on it. This can break downstream systems where now, if `cabal2nix` sees a .cabal file it removes
```
libraryToolDepends = [ hpack ];
prePatch = "hpack";
```
from the generated derivation and then those derivations break in an environment where .cabal files are not present (eg. CI).

pwm:
@sternenseemann first of all thank you for your amazing work on all of haskell-nix! My comment comment is mostly for anyone else stumbling on this issue. Nevertheless I think flagging these as you say in the changelog would be good, if possible.

PR #509, opened 2021-07-07
liquid-fixpoint: Remove ocaml dependency

nh2:
It is no longer used since 2015, see:

* https://github.com/ucsd-progsys/liquid-fixpoint/commit/a6bfa04a1d922941fbba860f9f55c4f2ae18f9b6
* https://github.com/ucsd-progsys/liquid-fixpoint/commit/412c3e95c0f606b58a52cd8a69c62ba9b4902668


cdepillabout:
This looks reasonable, thanks!

PR #513, opened 2021-07-21
Cabal2nix 2.18.0

sternenseemann:


sternenseemann:
Obviously this is towards a release of `cabal2nix` at current `master`. As far as I can tell we have no breaking changes since the last release, but some nice bug fixes which is why I think doing this minor release is worth it (we can even backport it to NixOS 21.05!).

Please do check if I missed something or something looks odd. Also there's a [Package Candidate on Hackage](https://hackage.haskell.org/package/cabal2nix-2.17.1/candidate) for checking for any additional mistakes.

cdepillabout (approved):
This all looks good to me!

Thanks for continuing to work on this!

sternenseemann:
I realized that two PRs included in the release change the API of the exposed library although they don't really cause breaking changes for our actual users (marked the changes in the changelog). Still we should follow PVP and increase to 2.18.0.

Waiting for someone to sign off on this and then I'll merge and release.

maralorn (approved):
LGTM, thanks!

PR #520, opened 2021-09-22
cabal2nix: Assume "unknown" hackage license to be free

maralorn:
We are currently very inconsequent in regard to "unknown" licenses. In
general they stem from packages having their license field set to
"LicenseRef-OtherLicense" accompanied by a LICENSE file which specifies
the real license.

Currently in nixpkgs "unknown" licenses result in hydra builds being
disabled for that package, but the package is not treated as "unfree" by
anything else in nixpkgs. Users don‘t get warnings when they use them
and we actually very often build those packages on hydra because they
are dependencies on other packages with free licenses.

This change simply acknowledges the fact that we are already building
and distributing most of these packages by giving them their own hydra
job.

Here are the arguments in favor of this:

* We have been distributing those packages all along and no one complained.
* hackage rules require all packages to be "open source in spirit".
* There are no known unfree packages on hackage.
* If we notice an unfree package we can always declare the package unfree in the
hackage2nix config.

This is the easiest solution to deal with the existing inconsistencies.
All other approaches would be more invasive and would mean more
maintainer overhead.

While this might not be the best overall solution for dealing with
licenses, e.g. we could try to get rid of all stringly licenses, we can
always do that improvements later.

### DIFF
--- a/src/Distribution/Nixpkgs/Haskell/FromCabal/License.hs
+++ b/src/Distribution/Nixpkgs/Haskell/FromCabal/License.hs
@@ -163,7 +163,17 @@ fromSPDXLicense (SPDX.License expr) =
       -- Use the SPDX expression as a free-form license string.
       Unknown (Just $ prettyShow expr)
 
+-- "isFreeLicense" is used to determine whether we generate a "hydraPlatforms =
+-- none" in the hackage2nix output for a package with the given license.
+
+-- Note: If "isFreeLicense" returned false for a license which is not an unfree
+-- license from "lib.licenses" the package would still be build by hydra if
+-- another package depended on it.
+
+-- Since all software on hackage needs to be "open source in spirit" and we
+-- don‘t know any software on hackage for which we are not allowed to
+-- distribute binary outputs we assume that a package has a free license if we
+-- don‘t explicitly know otherwise.
 isFreeLicense :: Distribution.Nixpkgs.License.License -> Bool
 isFreeLicense (Known "lib.licenses.unfree") = False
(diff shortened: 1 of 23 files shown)


sternenseemann:
Crucially, hackage doesn't require uploaders to grant anyone permission to distribute binary forms of their packages. However in practice this is probably not an issue. The restrictions on binary redistributions are usually accompanied with a requirement to distribute the source code with it (arguably we comply to this since you can trivially obtain the `src` attributes uses) and/or the ability to relink the product with different libraries which we allow for via overriding etc.

Also as you've argued in effect this doesn't really change anything.

sternenseemann:
As a side note, we should probably add a way to blacklist a package (which, I guess replaces the package with a `throw`), that'd also be useful for the spam category on hackage.

src/Distribution/Nixpkgs/Haskell/FromCabal/License.hs
```diff
@@ -165,5 +165,5 @@ fromSPDXLicense (SPDX.License expr) =
 
 isFreeLicense :: Distribution.Nixpkgs.License.License -> Bool
 isFreeLicense (Known "lib.licenses.unfree") = False
-isFreeLicense (Unknown Nothing)             = False
+isFreeLicense (Unknown Nothing)             = True
```
cdepillabout, line 168: Could you leave a comment on this line explaining the reasoning for this?

Probably just copying some of the PR comment body here would be fine.

If someone is looking through this code, I imagine they might be surprised that unknown licenses are marked as free.

Other than that, this LGTM!

PR #521, opened 2021-09-28
Distribution.Nixpkgs.Haskell.FromCabal: no longer emit dep on webkitgtk24x

sternenseemann:
webkitgtk24x-gtk{2,3} have been removed from nixpkgs a while ago and
only continue to exist as aliases which throw, informing about the
removal. This is informative, but depends on evaluating with aliases
enabled which may not always be the case.

Instead we just remove this now essentially useless code from
cabal2nix. This will usually cause a dependency on the current version
of webkitgtk and the build to fail -- then we can mark the resulting
package as broken normally which was prevented before by eval
oddities (I presume at least).


maralorn (approved):
Looks very reasonable.

But maybe this should warrant a CHANGELOG entry and a major version bump? Not sure …

sternenseemann:
We will do a major version bump anyways and a changelog entry never hurts, you're right. Care to proofread?

maralorn:
Great!

PR #528, opened 2021-11-05
Documentation changes to cope with haskell.lib.compose

expipiplus1:
`haskell.lib` is used in cabal2nix itself, but since those functions
aren't being deprecated yet it would be a bit breakful to replace that
right now.

### DIFF
--- a/doc/frequently-asked-questions.rst
+++ b/doc/frequently-asked-questions.rst
@@ -497,24 +497,29 @@ The solution is to add it as another package to the environment:
 Quality assurance
 ~~~~~~~~~~~~~~~~~
 
-The ``haskell.lib`` library includes a number of functions for checking
-for various imperfections in Haskell packages. It’s useful to apply
-these functions to your own Haskell packages and integrate that in a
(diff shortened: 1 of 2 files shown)


doc/frequently-asked-questions.rst
```diff
@@ -497,24 +497,29 @@ The solution is to add it as another package to the environment:
...
-to assure your packages maintain a minimum level of quality. This
-section discusses some of these functions.
+The ``haskell.lib.compose`` library, (herein referred to as ``haskellLib``)
+includes a number of functions for checking for various imperfections in
+Haskell packages. It’s useful to apply these functions to your own Haskell
+packages and integrate that in a Continuous Integration server like `hydra
+<https://nixos.org/hydra/>`__ to assure your packages maintain a minimum level
+of quality. This section discusses some of these functions.
+
+There also exists ``haskell.lib`` with the same functionality, but with a less
```
sternenseemann, line 507:
```suggestion
``haskell.lib`` also offers the same functionality, but with a less
```

sternenseemann:
Oh we haven't merged this yet?

PR #543, opened 2022-03-04
shell.nix: Set up ghc environment, add haskell-language-server

Profpatsch:
Manually lists the nix dependencies required by cabal2nix; this of
course requires that the local <nixpkgs> version agrees with the
package versions of cabal2nix dependencies, so ymmv.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,7 @@
+if type -p lorri; then
+    echo "direnv: found lorri, using" 1>&2
+    eval "$(lorri direnv)"
+else
+    echo "direnv: using direnv nix support" 1>&2
+    use nix;
+fi
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,31 @@
+{ pkgs ? import <nixpkgs> {} }:
+
+let
+  haskellPackages = pkgs.haskellPackages;
+  ghc = haskellPackages.ghcWithHoogle (hps: [
+    hps.ansi-wl-pprint
+    hps.distribution-nixpkgs
+    hps.hackage-db
+    hps.hopenssl
+    hps.hpack
+    hps.language-nix
+    hps.lens
+    hps.optparse-applicative
+    hps.pretty
+    hps.split
+    hps.yaml
+    hps.monad-par
+    hps.monad-par-extras
+    hps.tasty
+    hps.tasty-golden
+  ]);
+
+in pkgs.stdenv.mkDerivation {
+  name = "shell";
+  buildInputs = [
+            ghc
+            haskellPackages.cabal-install
+            haskellPackages.haskell-language-server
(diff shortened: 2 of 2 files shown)


maralorn (approved):
Works for me and looks sane.

I actually think peti had no shell.nix on purpose. So what are the downsides?
Could get out of sync? Still think if someone needs a shell.nix having a broken shell.nix to start of is probably better than nothing?

What do the others think?

maralorn:
Have thought about this a bit more.

I wonder: Is it really necessary to not use `developPackage` or `shellFor` here? I mean we have come this far without the shell.nix so it wouldn‘t actually produce a boostraping issue.

If we don‘t do it we should add a comment or give a reason why this `shell.nix` is the way it is. I am a bit worried that someone might think that this shell.nix could be a good example to copy and modify.

sternenseemann:
Just checking in a cabal2nix-generated file would be even less of a maintenance hassle and evenly unproblematic in terms of bootstrapping.

Profpatsch:
Ah, I did the simplest possible thing to get it to work, didn’t want to think about bootstrapping so I avoided using the cabal2nix stuff from nixpkgs. otoh it’s already in there, so no harm done.

Profpatsch:
> I actually think peti had no shell.nix on purpose. So what are the downsides?

Pretty sure peti always relied on cabal itself downloading dependencies. I can see why that is good, since somebody’s `<nixpkgs>` might not be the right one to have all the right versions. idk

otoh we can’t really get haskell-language-server to work if we rely on cabal.

Profpatsch:
The only goal here was to be able to use haskell-language-server to do some refactorings.

maralorn:
The autogenerated shell.nix does not contain hls, but instead a version number which will get outdated even faster.

I suggest merging as is. (Maybe modulo an explaining comment.)

Profpatsch:
> The autogenerated shell.nix does not contain hls, but instead a version number which will get outdated even faster.

iirc it usually also contains `src = ./.` which is pretty bad for local development.

alexfmpe:
This had me confused for a few minutes as adding a dependency to .cabal wouldn't be reflected in the nix shell

> I wonder: Is it really necessary to not use developPackage or shellFor here? I mean we have come this far without the shell.nix so it wouldn‘t actually produce a boostraping issue.

Right, we're not removing ability, only adding convenience. 
Does local dev even affect bootstrapping? 

> Just checking in a cabal2nix-generated file would be even less of a maintenance hassle and evenly unproblematic in terms of bootstrapping.

We could go even further and check in an attribute for a bootstrapping 'cabal2nix' from a pinned nixpkgs, since that one would be built based on an (earlier) cabal2nix 's output when ran on itself?

PR #544, opened 2022-03-04
Refactors (ApplicativeDo, record syntax, fetchWith)

Profpatsch:
These are a few code refactors, 

* use `ApplicativeDo` where possible
* use explicity record syntax instead of ordering-dependent constructors
* reduce complexity of `fetchWith` by moving most arguments into a single enum in multiple steps

### DIFF
--- a/src/Distribution/Nixpkgs/Fetch.hs
+++ b/src/Distribution/Nixpkgs/Fetch.hs
@@ -2,11 +2,14 @@
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE LambdaCase #-}
 
 module Distribution.Nixpkgs.Fetch
   ( Source(..)
   , Hash(..)
   , DerivationSource(..), fromDerivationSource, urlDerivationSource
+  , DerivKind(..)
+  , derivKindFunction
   , fetch
   , fetchWith
   ) where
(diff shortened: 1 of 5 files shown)


src/Distribution/Nixpkgs/Fetch.hs
```diff
@@ -211,11 +219,6 @@ fetchWith (supportsRev, kind, command, addArgs) source = do
                  Right ds -> return (Just (ds { derivKind = kind }, BS.unpack l))
  where
```
maralorn, line 270: You accidentally produced an empty trailing `where`.

src/Distribution/Nixpkgs/Fetch.hs
```diff
@@ -115,7 +117,7 @@ fromDerivationSource DerivationSource{..} =
 
 -- | Fetch a source, trying any of the various nix-prefetch-* scripts.
 fetch :: forall a.
-         Bool                                   -- ^ If True, fetch submodules when the source is a git repository
+         FetchSubmodules                                   -- ^ whether to fetch submodules when the source is a git repository
```
maralorn, line 120: Looks like the comment is not aligned anymore.

maralorn (approved):
This PR is a delight. Thank you very much!

Just have two tiny nitpicks.

I’ll give others a chance to chime in as well. But from my point of view this is basically good to go.

Just out of curiosity: What was your motivation for this PR?

sternenseemann:
(force-pushed to trigger CI with 8.8.4)

Profpatsch:
> Just out of curiosity: What was your motivation for this PR?

I … accidentally looked at the code and had a bunch of time :sweat_smile:

Profpatsch:
I had some weird rebasing issues, but now it should be rebased on master & the comments fixed

maralorn:
Okay, thanks!

PR #546, opened 2022-04-20
libNixName: Add nix-cmd pkg-config module

roberth:
A hercules-ci-cnix-expr dependency moved into this lib and I had to add an override to make it work.

Specifying it here is supposedly cleaner, but can we move this mapping into Nixpkgs instead? It doesn't seem to be Haskell-specific at all.
Anyway, would be practical to get this in for the time being.


sternenseemann:
Feel free to create a nixpkgs PR updating `cabal2nix-unstable`. I want to get to a cabal2nix release eventually, but I've been stuck a little since a few months.

PR #547, opened 2022-04-25
PostProcess: Use lua5_4 for lua

leungbk:
The latest version of haskellPackages.lua uses lua5_4.

hslua does not directly depend on pkgs.lua any longer, so it is safe
to remove the hook for the newest hslua at the time of
writing (version 2.2.0).


src/Distribution/Nixpkgs/Haskell/FromCabal/PostProcess.hs
```diff
@@ -129,7 +129,7 @@ hooks =
   , ("hoogle", set testTarget "--test-option=--no-net")
   , ("hsignal < 0.2.7.4", set phaseOverrides "prePatch = \"rm -v Setup.lhs\";") -- https://github.com/amcphail/hsignal/issues/1
   , ("hslua < 0.9.3", over (libraryDepends . system) (replace (pkg "lua") (pkg "lua5_1")))
-  , ("hslua >= 0.9.3 && < 2.0.0", over (libraryDepends . system) (replace (pkg "lua") (pkg "lua5_3")))
+  , ("hslua >= 0.9.3 && < 2.2.0", over (libraryDepends . system) (replace (pkg "lua") (pkg "lua5_3")))
```
cdepillabout, line 132: Looking at the cabal file for `hslua-2.0.0` and `hslua-2.1.0`, it looks like there are no system packages needed.  Is this change here correct?

For example: https://hackage.haskell.org/package/hslua-2.0.0/hslua.cabal

cdepillabout, line 132: Looking at the cabal file for `hslua-2.0.0` and `hslua-2.1.0`, it looks like there are no system packages needed.  Is this change here correct?

For example: https://hackage.haskell.org/package/hslua-2.0.0/hslua.cabal

src/Distribution/Nixpkgs/Haskell/FromCabal/PostProcess.hs
```diff
@@ -142,7 +142,8 @@ hooks =
   , ("libxml", set (configureFlags . contains "--extra-include-dir=${libxml2.dev}/include/libxml2") True)
   , ("liquid-fixpoint", set (testDepends . system . contains (pkg "z3")) True . set (testDepends . system . contains (pkg "nettools")) True . set (testDepends . system . contains (pkg "git")) True . set doCheck False)
   , ("liquidhaskell", set (testDepends . system . contains (pkg "z3")) True)
-  , ("lua >= 2.0.0", over (libraryDepends . system) (replace (pkg "lua") (pkg "lua5_3")))
+  , ("lua >= 2.0.0 && < 2.2.0", over (libraryDepends . system) (replace (pkg "lua") (pkg "lua5_3")))
+  , ("lua >= 2.2.0 && < 3.0.0", over (libraryDepends . system) (replace (pkg "lua") (pkg "lua5_4")))
```
cdepillabout, line 146: Why the upper bound on `3.0.0` here?

cdepillabout, line 146: Why the upper bound on `3.0.0` here?

src/Distribution/Nixpkgs/Haskell/FromCabal/PostProcess.hs
leungbk: You're right, fixed.

leungbk: You're right, fixed.

src/Distribution/Nixpkgs/Haskell/FromCabal/PostProcess.hs
leungbk: OK, I've removed it.

leungbk: OK, I've removed it.

cdepillabout:
LGTM, thanks!

For anyone interested, here is the changelog for the haskell package lua about 2.2.0 bumping up to lua-5.4: https://hackage.haskell.org/package/lua-2.2.0/changelog

PR #549, opened 2022-04-27
Flags: remove unnecessary "https" flag from Pandoc

leungbk:
See https://github.com/jgm/pandoc/commit/d414b2543a1686007e84c54bc711dff969dfb569.


cdepillabout:
This looks reasonable.  Recent versions of pandoc don't seem to have this flag anymore (as per the linked commit).  For example:

https://hackage.haskell.org/package/pandoc-2.18

Thanks, LGTM!

PR #552, opened 2022-05-14
Convert "gcc_s" dependency to a configuration flag

### LINKED ISSUE #71: Deal with gcc_s in extra-libraries field
The `bits-extra` package has `gcc_s` in extra-libraries field. What `cabal2nix` does is put `gcc_s` as an extra argument and makes `extraLibraries = [ gcc_s ];`, which doesn't work because such package doesn't exist. I am told that it instead needs to do something with `NIX_LDFLAGS="-lgcc_s"` and have `gcc` as a dependency.

leungbk:
Fixes #71.


sternenseemann:
Where are you running into this? Merely two packages declare a dependency on `gcc_s` in all of Hackage these days and both are broken for unrelated reasons.

PR #555, opened 2022-05-22
Distribution.Nixpkgs.Haskell.Platform: unify CLI platform parsing

sternenseemann:
In the cabal2nix project there are two flags that accept some kind of
string that needs to be converted to a `Platform` which is used for
evaluating a `.cabal` file:

* cabal2nix: `--system` accepts `cpu-vendor-os` and `cpu-os`
* hackage2nix: `--platform` only accepts `cpu-os`.

Their names should be swapped, ironically compared to what they do which
should be tackled eventually.

Initially I thought the parsing of these could be replaced by
distribution-nixpkgs new capabilities, but ran into the following issue:
distribution-nixpkgs needs to be lossless, since `Platform`s need to be
rendered again.

This is problematic when evaluating Cabal files, however: Cabal has no
notion of powerpc64le and powerpc64 and GHC's conversion logic causes
them to be both interpreted as PPC64. Package authors likely expect this
behavior, so we risk accidentally getting wrong results by insisting on
losslessness.

To tackle this, we'll have two different ways to parse a
nixpkgs-originating string into a `Platform`:

* One in distribution-nixpkgs to parse into a `NixpkgsPlatformSingle`
* One in cabal2nix to parse into a genuine Cabal `Platform`

### DIFF
--- a/src/Distribution/Nixpkgs/Haskell/Platform.hs
+++ b/src/Distribution/Nixpkgs/Haskell/Platform.hs
@@ -1,0 +1,123 @@
+{-# LANGUAGE LambdaCase #-}
+{-|
+Description: Parse platform strings used by nixpkgs into Cabal's 'Platform'
+
+This module defines conversions from the (autoconf-derived) platform strings
+nixpkgs uses into Cabal's 'Platform' type. This is intended to facilitate later
+evaluation of @.cabal@ files. For this conversion Cabal's 'Permissive'
+heuristics are used as well as a logic equivalent to the @GHC_CONVERT_*@ macros
+from GHC's configure script.
+
+Since the process is inherently lossy because Cabal ignores certain factors like
+endianness, conversion from 'Platform' to nixpkgs' platform strings. For this
+usecase, try "Distribution.Nixpkgs.Meta" from @distribution-nixpkgs@.
+-}
+module Distribution.Nixpkgs.Haskell.Platform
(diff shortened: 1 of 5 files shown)


src/Distribution/Nixpkgs/Haskell/Platform.hs
```diff
@@ -0,0 +1,123 @@
...
+--   'parsePlatformLenient'.
+--
+--   >>> parsePlatformFromSystem "x86_64-linux"
+--   Just (Platform X86_64 Linux)
+--   >>> parsePlatformFromSystem "x86_64-linux-musl"
+--   Just (Platform X86_64 Linux)
+--   >>> parsePlatformFromSystem "i686-netbsd"
+--   Just (Platform I386 NetBSD)
+parsePlatformFromSystem :: String -> Maybe Platform
+parsePlatformFromSystem s =
```
Profpatsch, line 117: If it has the same caveats as `parsePlatformLenient`, I’d call it `parsePlatformFromSystemLenient`

Profpatsch (approved):
I haven’t checked the parsing logic in-depth, but looks good.

src/Distribution/Nixpkgs/Haskell/Platform.hs
sternenseemann: Thanks, that's true.

Issue #504, opened 2021-06-12
RFC: move postprocessing into overlays in nixpkgs as much as possible?


sternenseemann:
`src/Distribution/Nixpkgs/Haskell/FromCabal/PostProcess.hs` has a lot of post processing hooks which do things that would also be possible via overrides, like disabling tests, adding dependencies, adding a phase hook etc.

I've stumbled over this before since cabal2nix sometimes does “magic” stuff I didn't expect. To me it'd be clearer to move these changes into `nixpkgs`'s `configuration-*.nix` files, but there is the risk of breaking stuff for users of `callCabal2nix`, `callHackageDirect` etc.

maralorn:
I think we should strive at being very clear cut here.

For every kind of override we do we should decide whether to do it in cabal2nix or in nixpkgs. And this distinction should be clearly documented to minimize the danger of confusion.

But I exactly agree with your analysis, that I would prefer to have as much as possible in nixpkgs and am not sure what would be optimal. Maybe a good intuition is this: Stuff that would fit into configuration-common.nix would probably not fit callCabal2nix and callHackage. But stuff that would go into configuration-nix.nix would be nice if it also applied to them (not so much for callCabal2Nix but definitely for callHackage. If we can somehow figure this out we can probably spare people a few overrides.) But how can we achieve this?

I am brainstorming for ideas here:

* A possible middle ground would be having a config file for this which we maintain in nixpkgs but which will also be used by callCabal2nix and callHackage.

* Another possibility would be to expose an override option on haskellPackages where people can somehow substitute packages _before_ the configuration-*.nix (or at least configuration-nix.nix) overrides get applied.

* We could theoretically somehow bake the overrides into mkDerivation? (With an option to disable them?)  (I know this sounds kinda crazy but I actually think this could be more flexible if done right.

Having a "INFO: cabal2nix has applied the following auto-detected fixes to this derivation" somewhere in the build log would be sweet.

cdepillabout:
> I think we should strive at being very clear cut here. For every kind of override we do we should decide whether to do it in cabal2nix or in nixpkgs. And this distinction should be clearly documented to minimize the danger of confusion.

Regardless of what we eventually decide to do about this, I imagine this is the most important thing.

-------------------------------------

I don't personally have a strong feeling about what to do about this, but I did want to bring up a few other points:

[...]

PR #556, opened 2022-05-22
pandoc: install bash completions

leungbk:


sternenseemann:
It's probably more maintenable to put these kind of things into nixpkgs in the future. Ref. #504

src/Distribution/Nixpkgs/Haskell/FromCabal/PostProcess.hs
```diff
@@ -396,6 +396,8 @@ pandocOverrides = set phaseOverrides postInstall
     postInstall = unlines [ "postInstall = ''"
                           , "  mkdir -p $out/share/man/man1"
                           , "  mv \"man/\"*.1 $out/share/man/man1/"
+                          , "  mkdir -p $out/share/bash-completion/completions"
+                          , "  mv data/bash_completion.tpl $out/share/bash-completion/completions/pandoc"
```
sternenseemann, line 400: Should probably use `pandoc --bash-completion` and thus needs to be tied to a check whether we are cross compiling, that's easier in nixpkgs.

leungbk:
OK, I'll submit a PR to nixpkgs modifying `configuration-common.nix` instead.

sternenseemann:
> OK, I'll submit a PR to nixpkgs modifying `configuration-common.nix` instead.

`configuration-nix.nix` for this sort of thing, please, but sounds good overall!

PR #557, opened 2022-06-14
Distribution.Nixpkgs.Haskell.FromCabal: mainProgram check buildable

sternenseemann:
Naturally only a buildable executable can be the mainProgram. Checking
this property eliminates some false positive mainPrograms and also gets
rid of some cases where it wasn't recognized properly. pandoc still
eludes cabal2nix, probably something strange going on with flag
resolution.

### DIFF
--- a/src/Distribution/Nixpkgs/Haskell/FromCabal.hs
+++ b/src/Distribution/Nixpkgs/Haskell/FromCabal.hs
@@ -149,7 +149,7 @@ fromPackageDescription haskellResolver nixpkgsResolver missingDeps flags Package
     -- it is hard to decide automatically which should be the default/main one.
     nixMainProgram :: Maybe String
     nixMainProgram =
-      case executables of
+      case filter (buildable . buildInfo) executables of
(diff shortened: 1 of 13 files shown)


maralorn:
This looks like it introduces a lot of false negatives? e.g. brittany certainly has a sensible mainProgram?

sternenseemann:
I think something weird happened when I tried to update the golden files, hang on.

sternenseemann:
Looks /much/ more sensible now!

maralorn:
It’s still unclear to me why there is a new mainProgram appearing if you introduce a filter, but I am sure it is fine.^^

sternenseemann:
> It’s still unclear to me why there is a new mainProgram appearing if you introduce a filter, but I am sure it is fine.^^

If two executables are declared, but only one is buildable, previously no mainProgram would be emitted, but now it would.

E.g. pandoc has `pandoc` and `trypandoc`, but only `pandoc` is buildable by default.

maralorn:
Uh, beautiful! Thx!

PR #559, opened 2022-06-19
Distribution.Nixpkgs.Fetch: Remove unnecessary DerivKindZip

sternenseemann:
17d47044b34f891fe3ff4c5d05e5855dd0ae8c8a elaborated on DerivKindUrl and
DerivKindZip in the name of saving an argument list, but accidentally
created two new cases that were unreachable:

* DerivKindUrl UnpackArchive
* DerivKindZip DontUnpackArchive

Both wouldn't have worked in practice as well, since they would use the
correct nix-prefetch-script, but the wrong nixpkgs fetcher (since
derivKindFunction ignored the extra parameter).

To get to the two actual cases again to the four we accidentally
introduced, we can simply remove DerivKindZip and adjust
derivKindFunction to look at the parameter:

* DerivKindUrl UnpackArchive: nix-prefetch-url --unpack, fetchzip
* DerivKindUrl DontUnpackArchive: nix-prefetch-url, fetchurl

### DIFF
--- a/src/Distribution/Nixpkgs/Fetch.hs
+++ b/src/Distribution/Nixpkgs/Fetch.hs
@@ -131,7 +131,7 @@ fetch optSubModules f = runMaybeT . fetchers where
   fetchers :: Source -> MaybeT IO (DerivationSource, a)
   fetchers source = msum . (fetchLocal source :) $ map (\fetcher -> fetchWith fetcher source >>= process)
     [ (False, DerivKindUrl DontUnpackArchive)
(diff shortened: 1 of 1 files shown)


Profpatsch (approved):

maralorn (approved):
I am tempted to bike shed whether, I like to use the name `DeriveKindUrl` for `fetchzip` But nix-prefetch-url does the same thing, so we are fine, I guess.

This looks reasonable. Thank you!

PR #575, opened 2022-10-10
hackage-db: Fix the type of signatures in metadata

### LINKED ISSUE #573: `signatures` field in `MetaData` has the wrong type

sternenseemann:
Previous PR/discussion: https://github.com/NixOS/hackage-db/pull/17

---

Signatures are actually complex objects. We could use `Value` here since it's unlikely that any consumers care about this, but it's cheap enough to just define a structure for the signature objects.

In practice this isn't an issue because it seems like this field is never populated in Hackage, but it can be populated if you build a custom Hackage it can happen.

Fixes #573.

cc @michaelpj (sorry for the repository moving mess)


michaelpj:
Thanks!

PR #576, opened 2022-10-10
cabal2nix: resolve crypt library to libxcrypt

sternenseemann:
NixOS has disabled glibc's libcrypt and is using libxcrypt now (starting with 22.11, that is): https://github.com/NixOS/nixpkgs/pull/181764

This means we'll need to resolve the crypt identifier to libxcrypt. Releases including this commit need to go into NixOS unstable (as soon as the current staging iteration has finished) and NixOS 22.11. A release including this commit *may not* be backported to 22.05, as it'd likely break builds there.


cdepillabout (approved):
This looks reasonable to me.

Great that one of our golden tests covered this as well.

sternenseemann:
Tested on current `haskell-updates` (the libxcrypt changes got merged), seems to work as intended. We should also make a release before 22.11.

PR #577, opened 2022-10-12
distribution-nixpkgs: prepare 1.7.0.1 release

sternenseemann:
This release is mostly for the benefit of `haskell.packages.ghc942`, since we currently need to disable the test suite there.


cdepillabout (approved):
LGTM!

PR #579, opened 2022-10-13
ormolu/fourmolu: workaround probable macOS linker bug

lf-:
https://github.com/tweag/ormolu/issues/927 is a bug where fourmolu/ormolu cannot build on macOS due to using TH to build some massive value, which then segfaults the linker.

It has a flag to Not Do That and instead embed the file, so this PR enables that flag.


cdepillabout:
Enabling this flag would in theory make ormolu/fourmolu slightly slower, right?

I'd prefer not to do this in `cabal2nix`, but instead just add it to `pkgs/development/haskell-modules/configuration-darwin.nix` until the GHC bug (?) is fixed.

lf-:
Hmmm. I thought that was impossible since it would have to apply to hackage2nix to add dependencies as it needs to, but I think the dependency could be added by overrideCabal instead

sternenseemann:
> I'd prefer not to do this in `cabal2nix`, but instead just add it to `pkgs/development/haskell-modules/configuration-darwin.nix` until the GHC bug (?) is fixed.

Agreed.

lf-:
Thanks for the advice. It's been dealt with in nixpkgs.

PR #581, opened 2022-10-22
cabal2nix: bump to 2.19.1 and add changelog

sternenseemann:


cabal2nix/CHANGELOG.md
```diff
@@ -1,5 +1,30 @@
 # Revision History for cabal2nix
 
+## 2.19.1
+
+**Warning**: This version of `cabal2nix` generates Nix expressions that
+are expected to be used with at least NixOS 22.11 or the `nixos-unstable`
+channel at 22.05-pre-86574-g3928cfa27d9 or later due to changes how
```
cdepillabout, line 7: Maybe give an approximate date here, just so users have some idea of when this commit hash refers to?

cdepillabout (approved):
LGTM, thanks for writing good changelog summaries!

cabal2nix/CHANGELOG.md
sternenseemann: I was meaning to, but I don't think there's a channel released with the change yet. Also date is bound to be confusing with the different unstable channel variants, so it's better if ppl check `nixos-version` instead of comparing dates if something is not working. In general I hope this note will just make everyone check that they have an up to date channel and that's that.  For debugging issues the hash is better in any case.

I do need to figure out why it says `22.05-pre` though, that seems wrong.

cdepillabout: I imagine for an end user, looking at a date is likely much easier than actually checking a hash (even if it is an imperfect heuristic).

maralorn (approved):
:tada:

PR #585, opened 2022-11-08
Fetch `.tar.gz` using `nix-prefetch-url --unpack`

pwm:
Fetching `tar.gz` files is common enough of a use-case that warrants treating them specially by bypassing other fetching options and using `nix-prefetch-url --unpack` directly. This speeds up large tarballs as `cabal2nix` currently downloads them twice as the first fetch assumes an unpacked url and thus will fail.


dschrempf (changes requested):
I am not a regular cabal2nix contributor, but I saw your discussion on the matrix channel. Here are my two cents:

I think the `fetchPackedUrl` function needs a top level comment explaining why it was added.

Aslo, I think the `process` function should be top-leveled and reused (by `fetch` and `fetchPackedUrl`).

pwm:
@dschrempf thanks for the feedback

> I think the fetchPackedUrl function needs a top level comment explaining why it was added.

Agreed, will add.

> Aslo, I think the process function should be top-leveled and reused (by fetch and fetchPackedUrl).

I actually did that first but then thought it might be better to try and minimise the change which is why i did it this way. Happy to lift it out (I think it makes sense) but that will mean a slightly bigger PR (not sure if that's an issue though).

PR #587, opened 2022-12-01
Flake-enable project

Gabriella439:
The immediate motivation for doing this was that
the old `shell.nix` failed with the `nixpkgs` channel that I had installed.  While I was fixing this to pin `nixpkgs` I figured this would be a good time to flake-enable the project.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,87 @@
+{ inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/release-22.11";
+
+    flake-compat = {
+      url = "github:edolstra/flake-compat";
+      flake = false;
+    };
+
+    flake-utils.url = "github:numtide/flake-utils/v1.0.0";
+  };
+
+  outputs = { nixpkgs, flake-utils, ... }:
+    flake-utils.lib.eachDefaultSystem (system:
+      let
+        ghcVersion = "ghc90";
+
+        config = { };
+
+        overlay = self: super: {
+          haskell = super.haskell // {
+            packages = super.haskell.packages // {
+              "${ghcVersion}" = super.haskell.packages."${ghcVersion}".override (old: {
+                overrides =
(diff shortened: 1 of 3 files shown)


sternenseemann:
This was proposed before in #533 and I'm inclined to veto it – I do not want to support anyone declaring `cabal2nix` as a flake input (which seems to be the inevitable consequence).

I also personally don't use flakes, so the sluggishness afforded by flake-compat is not really acceptable.

I'm curious about the eval failure though, since cabal2nix is pretty boring (which is also a reason why having some random channel over a months old pin is not bad).

flake.nix
```diff
@@ -0,0 +1,87 @@
...
+                  self.lib.composeExtensions
+                    (old.overrides or (_: _: { }))
+                    (hself: hsuper: {
+                      # The reason we stick these underneath a `local` attribute
+                      # is to avoid infinite recursion (because `callCabal2nix`
+                      # depends on all of these packages.
+                      local = rec {
+                        cabal2nix =
+                          hself.callCabal2nix "cabal2nix" ./cabal2nix {
+                            inherit distribution-nixpkgs hackage-db;
```
sternenseemann, line 33: This would also depend on `language-nix`.

```suggestion
                            inherit distribution-nixpkgs hackage-db language-nix;
```

Gabriella439:
If the requirement is that you want the `nixpkgs` input to remain up-to-date that is totally doable using flakes.  I can change the flake input to track Nixpkgs `master`, `nixpkgs-unstable`, or whatever branch you want

Gabriella439:
Also, just to clarify: the motivation for this flake is for local development, not for depending on `cabal2nix` as a flake input.  If you want to remove the possibility of depending on `cabal2nix` as a package I can remove the `packages` outputs from the flake and leave behind just the `devShells`.  I only included the `packages` outputs for completeness

PR #597, opened 2023-03-10
hackage-db: support cabal-install 3.10.1.0 XDG paths.

athas:
From 3.10.1.0 onwards, cabal-install uses XDG directories to store its state and cache components.  cabal2nix expects to find the Hackage tarball in ~/.cabal, where it may no longer be located.

This commit changes cabal2nix to look in ~/.cabal if that directory exists (which matches the behaviour of cabal-install itself), but otherwise looks in the appropriate XDG directory.

### DIFF
--- a/hackage-db/src/Distribution/Hackage/DB/Path.hs
+++ b/hackage-db/src/Distribution/Hackage/DB/Path.hs
@@ -17,12 +17,23 @@ import Control.Exception
 import System.Directory
 import System.FilePath
 
--- Some logic to handle both old ~/.cabal and XDG cache directory used
--- by default from cabal-install 3.10.1.0.
+-- |
+-- Determines the /state/ directory (which e.g. holds the hackage tarball)
+-- cabal-install uses via the following logic:
 --
--- This is a simplified form of the logic found in cabal-install
--- itself:
--- https://github.com/haskell/cabal/blob/0ed12188525335ac9759dc957d49979ab09382a1/cabal-install/src/Distribution/Client/Config.hs#L594-L610
+-- 1. If @~/.cabal@ (see 'getAppUserDataDirectory') exists, use that.
+-- 2. Otherwise, use @${XDG_CACHE_HOME}/cabal@ (see @'getXdgDirectory' 'XdgCache'@)
+--    which is the new directory cabal-install can use starting with
+--    version @3.10.*@.
+--
+-- This logic is mostly equivalent to what upstream cabal-install is
+-- [doing](https://github.com/haskell/cabal/blob/0ed12188525335ac9759dc957d49979ab09382a1/cabal-install/src/Distribution/Client/Config.hs#L594-L610)
+-- with the following exceptions:
+--
+-- * The @CABAL_DIR@ environment variable is not supported at the moment.
+-- * The state directory can freely be configured to use a different location
+--   in the cabal-install configuration file. hackage-db doesn't parse this
+--   configuration file, so differing state directories are ignored.
 cabalStateDir :: IO FilePath
 cabalStateDir = do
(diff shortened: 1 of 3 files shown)


sternenseemann:
Thanks! Can you also link to the [upstream code](https://github.com/haskell/cabal/blob/0ed12188525335ac9759dc957d49979ab09382a1/cabal-install/src/Distribution/Client/Config.hs#L594-L610) that implements this in the comment for reference?

This also shows that we should respect the `CABAL_DIR` environment variable, but we can also do that in a separate change, as we weren't even respecting it in the first place.

athas:
Link added.

The Right Solution is perhaps to depend on `cabal` to use exactly the same logic.  However, cabal2nix got away with not doing exactly what `cabal` does so far (in particular, not respecting `CABAL_CONFIG` or `CABAL_DIR`), so it probably does not matter in practice.

sternenseemann:
Seems like cabal-install has made our lives harder than it should be – namely you can also configure arbitrary location for the package db location. What's worse this is apparently rendered into the configuration file by default, so even if people move their config to xdg paths, it may still show up under `~/.cabal`.

Not *quite* sure yet how to proceed – we'll want to avoid depending on `cabal-install` I think, since it is quite the dependency. One nice thing is that we support a wide range of `Cabal` libraries, but supporting a wide range of `cabal-install` is likely very painful.

athas:
The issue with paths being stored in the cabal config file is not new.  It's been like that for a while - maybe forever.  This PR does not make the situation worse.

sternenseemann:
True, but it'll be more likely to be a problem now. In any case that we'll have to prefer `~/.cabal` anyways will get us out of 90% of the issues hopefully.

For anything else I'll implement `CABAL_DIR` and people will have to work around it manually.

hackage-db/src/Distribution/Hackage/DB/Path.hs
```diff
@@ -39,9 +50,11 @@ hackageTarballDir :: IO FilePath
 hackageTarballDir = cabalTarballDir "hackage.haskell.org"
 
 -- | Determine the default path of the Hackage database, which typically
--- resides at @"$HOME\/.cabal\/packages\/hackage.haskell.org\/00-index.tar"@.
--- Running the command @"cabal update"@ will keep that file up-to-date.
-
+-- resides in @$HOME\/.cabal\/packages\/hackage.haskell.org\/@.
```
maralorn, line 59: Do we want to adapt this path to the new default?

hackage-db/src/Distribution/Hackage/DB/Path.hs
```diff
@@ -15,10 +15,37 @@ import Distribution.Hackage.DB.Errors
...
+cabalStateDir = do
+  envCabal <- lookupEnv "CABAL_DIR"
+  dotCabal <- getAppUserDataDirectory "cabal"
+  dotCabalExists <- doesDirectoryExist dotCabal
+  case envCabal of
+    Just dir -> pure dir
+    Nothing ->
+      if dotCabalExists
+      then pure dotCabal
+      else getXdgDirectory XdgCache "cabal"
```
maralorn, line 48:
```suggestion
    Nothing | dotCabalExists -> pure dotCabal
    Nothing -> getXdgDirectory XdgCache "cabal"
```
Really unimportant. Just suggesting it to learn whether you also like it better and if not, why not.

maralorn (approved):
Excellent! I left to small comments, but can also be merged as is.

hackage-db/src/Distribution/Hackage/DB/Path.hs
athas: This was copy-pasted from the cabal source code which doesn't seem to use guards much.

PR #598, opened 2023-03-13
distribution-nixpkgs: match upstream rendering of JavaScript "arch"

sternenseemann:
[GHC] and following that [nixpkgs] now use the same string for the JavaScript arch as Cabal does, so we no longer need a special rule for that.

Update test data from nixpkgs and coverage tests:

- Reflected changed javascript-ghcjs double.
- The FreeBSD is now versioned by kernel version, so we use freebsd13 instead of freebsd. We'll handle this via OtherArch instead of updating the rendering of FreeBSD, since this eliminates the need to update this library as pkgs.freebsd is updated – additionally it is possible nixpkgs will support multiple FreeBSD versions at the same time in the future.
- Added:
  * mipsisa32r6-linux
  * mipsisa32r6el-linux
  * mipsisa64r6-linux
  * mipsisa64r6el-linux

[GHC]: https://gitlab.haskell.org/ghc/ghc/-/commit/6636b670233522f01d002c9b97827d00289dbf5c
[nixpkgs]: https://github.com/NixOS/nixpkgs/commit/471b9cab41b218080f5e9f4abbc83eaaa60c6abf


maralorn (approved):
LGTM, thanks!

PR #606, opened 2023-08-14
cabal2nix: resolve icu-i18n pkg-config module to icu

raehik:
The Nix package is icu, which provides three pkg-config modules including icu-i18n.


raehik:
Question: Are the existing ICU pkg-config module entries correct? [development/libraries/icu/base.nix](
https://github.com/NixOS/nixpkgs/blob/0b6d395909219e21fbd55dd6b321fc1f06a424ab/pkgs/development/libraries/icu/base.nix#L49-L53) states:

```nix
      pkgConfigModules = [
        "icu-i18n"
        "icu-io"
        "icu-uc"
      ];
```

So maybe we should swap out `icui18n`, `icuuc` etc. for `icu-i18n`, `icu-uc`. Or maybe we need both, if both are used?

sternenseemann:
Done in 0365d9b77086d26ca5197fb48019cedbb0dce5d2.

PR #616, opened 2024-01-31
Make hackage2nix locale independent

maralorn:
Since my system is currently missing the locale `C.UTF-8` which we set in `regenerate-hackage-packages.nix` I want to make `hackage2nix` locale independent. The impurity is driving me nuts. Since we all edit the same file with hackage2nix hardcoding the locale makes total sense.

From memory I think the locale can affect two things:

1. It affects the encoding with which hackage-packages.nix get’s written.
2. It affects the ordering of entries in hackage-packages.nix

After reading the documentation for `base:System.IO` I am moderately certain that this change fixes 1.

Regarding 2 I am just confused. From reading the code I think the ordering depends on the Ord instances of PackageName and Version. Those are newtypes which in the end all boil down to the Ord instances of ShortByteString or String or Int or Word64. I haven’t found any indication that any of them is locale dependent, but I might have overlooked something. Or the ordering issue is something completely else? I don’t know.

At least I think setting the locale explicitly this way won’t make things less reproducible.

### DIFF
--- a/cabal2nix/CHANGELOG.md
+++ b/cabal2nix/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Revision History for cabal2nix
 
+## Unreleased
+
+* `hackage2nix` now defaults internally to a utf-8 locale ignoring enviroment
+  variables.
+
 ## 2.19.1
 
 **Warning**: This version of `cabal2nix` generates Nix expressions that
--- a/cabal2nix/hackage2nix/Main.hs
+++ b/cabal2nix/hackage2nix/Main.hs
@@ -41,6 +41,7 @@ import Text.PrettyPrint.HughesPJClass hiding ( (<>) )
 import Data.List.NonEmpty (NonEmpty)
 import Data.Semigroup (sconcat)
 import Options.Applicative.NonEmpty (some1)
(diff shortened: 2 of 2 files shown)


maralorn:
I have tested this with and without "LC_ALL=C.UTF-8" on my system and the hackage-packages.nix file always is the same. (Without the fix in this PR with LC_ALL=C.UTF-8 would just fail.)

I suggest merging this PR and the dropping the LC_ALL line from the script in nixpkgs.

cdepillabout (approved):
This seems reasonable, thanks!

Might be nice to add a note in the CHANGELOG about this as well?

sternenseemann:
Does it actually affect the ordering of packages in `hackage-packages.nix` in practice, did you test that? I seriously doubt that was ever the case. IIRC the locale problems is only an issue due to the usage of `sort(1)` in the maintainer scripts.

maralorn:
Aaaah, yeah. I could not reproduce any change in ordering and I was highly confused because I also couldn’t imagine the Haskell sorting being unstable but I was sure I remembered some sorting problem. It makes total sense that this is about the .yaml files.

PR #619, opened 2024-05-07
distribution-nixpkgs: update test data + small cleanup

sternenseemann:


maralorn (approved):
Am AFK, so can't test, but this looks simple enough.

PR #625, opened 2024-06-22
Correct gnome-keyring mapping

jtojnar:
gnome-keyring is a daemon. The gnome-keyring package means to use libgnome-keyring.


maralorn:
Did you maybe base this change on an old checkout?

jtojnar:
Oops, fixed.

maralorn:
Thanks!

PR #629, opened 2024-10-05
Name: use explicit ABI version of webkitgtk

fabianhjr:
Context / Related to: https://github.com/NixOS/nixpkgs/pull/345611


sternenseemann:
I think the versions without a version specifier should always resolve to the generic attribute (which isn't being removed in nixpkgs, right?). We can still (semi-temporarily) pin it to a fixed version via `configuration-common.nix` for the affected packages in nixpkgs, but don't create an awkward situation where we generate references to 4.0 for downstream packages that actually use/support a later version.

fabianhjr:
>  but don't create an awkward situation where we generate references to 4.0 for downstream packages that actually use/support a later version

From my understanding the default version might be erroneous since it is not the webkitgtk version but the ABI version witch incompatibilities between them (gtk3 vs gtk4 between 4_0 and 6_0 and libsoup2 vs libsoup3 between 4_0 and 4_1/6_0)

And the idea would be to remove the default alias. :c

cabal2nix/src/Distribution/Nixpkgs/Haskell/FromCabal/Name.hs
```diff
@@ -185,10 +187,14 @@ libNixName "wayland-client"                     = return "wayland"
 libNixName "wayland-cursor"                     = return "wayland"
 libNixName "wayland-egl"                        = return "libGL"
 libNixName "wayland-server"                     = return "wayland"
-libNixName "webkit"                             = return "webkitgtk"
-libNixName "webkit2gtk"                         = return "webkitgtk"
```
fabianhjr, line 189: Would it be better to drop this two then?

maralorn:
Removing the default alias and adding versioned names is kinda pointless. This is not an API which hackage packages adhere to, this is a map we maintain to react to whatever hackage packages do. Are there cabal files on hackage which use those versioned attributes? My guess is no.

The purpose of this map is imo, that the derivation expects a system library as input under the name which we use in nixpkgs. I agree with sterni, unless the version is already contained in the name used in the cabal file we should override the version in nixpkgs.

maralorn:
@sternenseemann Just out of curiosity, do you know whether there actually are cabal files which use the versioned package names?

sternenseemann:
@maralorn <https://github.com/NixOS/nixpkgs/commit/212b040a816616345dd01d3e77094b2c342cd2bc>. Any serious project should, frankly. If they use pkg-config, they are forced to (I think).

maralorn:
Ah, okay. Great, thank you!

PR #635, opened 2024-12-04
FromCabal.Name: fix libsoup attribute names

jtojnar:
`pkgconfig-depends: libsoup-3.0` was being incorrectly resolved to `libsoup` attribute. Additionally, https://github.com/NixOS/nixpkgs/pull/360908 renamed `libsoup` attribute to `libsoup_2_4`.

Let’s use the correct package attribute names for all libsoup pkg-config files. All libsoup dependent Haskell-packages use `pkgconfig-depends` with one of those names:

https://hackage.haskell.org/package/gi-soup-3.0.3/src/gi-soup.cabal
https://hackage.haskell.org/package/gi-soup-2.4.29/src/gi-soup.cabal
https://hackage.haskell.org/package/spike-0.3/src/spike.cabal


maralorn:
Thank you!

PR #636, opened 2024-12-12
Terminal colors for output

9999years:
This adds terminal colors to `cabal2nix`'s output. This checks if `stderr` `isatty` so that output will not be colored if (for example) `cabal2nix`'s `stderr` is being piped to a file. It also respects the `$NO_COLOR` and `$FORCE_COLOR` environment variables for toggling the output coloring (see [no-color.org](https://no-color.org/)).

A couple small changes while I was preparing this:

1. I added `nix-prefetch-scripts` to the `shell.nix` so that you can run `cabal2nix` successfully while hacking on it.
2. I added a line to print the fetcher command (like `nix-prefetch-git ...`) before we run it. This makes it a lot clearer where the errors come from, and which errors are being ignored (because we're trying another fetcher).

Example output:
image

### DIFF
--- a/cabal2nix/src/Distribution/Nixpkgs/Fetch.hs
+++ b/cabal2nix/src/Distribution/Nixpkgs/Fetch.hs
@@ -25,12 +25,12 @@ import Data.Aeson
 import qualified Data.ByteString.Lazy.Char8 as BS
 import qualified Data.List as L
 import Data.Maybe
+import Distribution.Nixpkgs.Color (colorStderrLn, commandColor, warningColor)
 import GHC.Generics ( Generic )
 import Language.Nix.PrettyPrinting as PP
 import System.Directory
 import System.Environment
 import System.Exit
-import System.IO
 import System.Process
 
 -- | A source is a location from which we can fetch, such as a HTTP URL, a GIT URL, ....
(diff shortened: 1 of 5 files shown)


sternenseemann:
Good idea, I think this is a good migitation for the error message problem the current fetcher logic has!

Not quite sure what to think about colored output, i.e. if the extra code is worth it…

9999years:
@sternenseemann Happy to remove the colored output if you'd prefer (or you can update my branch -- just drop the last commit), let me know

sternenseemann:
I've merged those changes manually, we can leave this PR for the color changes. I'm going to keep thinking about it. Maybe @maralorn and @cdepillabout want to comment?

cdepillabout:
I think I'm generally in favor of (optional) colors.

For CLI programs that use color, I think I'd also expect a `--color` / `--no-color` CLI flag to exist, but I don't think not having it should necessarily think that should block this PR.

@9999years could you post a screenshot of what the colors look like in the output of cabal2nix?

9999years:
@cdepillabout Added a screenshot to the PR description!

cdepillabout:
I think the colors seems nice, and it will likely make visually parsing the output of `cabal2nix` easier.  I'm for merging this.

maralorn:
Unsurprisingly I am all in favor of colorful output. Code complexity seems manageable.

sternenseemann:
Alright. Would be good if someone could review and merge this PR soon-ish, I think releasing a new cabal2nix version is sort of overdue (and we should get out a fix for text >= 2.1.2 rather sooner than later).

9999years:
Anything needed for this?

cabal2nix/src/Distribution/Nixpkgs/Fetch.hs
```diff
@@ -219,7 +219,7 @@ derivKindFunction = \case
 fetchWith :: (Bool, DerivKind) -> Source -> MaybeT IO (DerivationSource, FilePath)
 fetchWith (supportsRev, kind) source = do
   unless ((sourceRevision source /= "") || isUnknown (sourceHash source) || not supportsRev) $
-    liftIO (hPutStrLn stderr "** need a revision for VCS when the hash is given. skipping.") >> mzero
+    colorStderrLn warningColor "** need a revision for VCS when the hash is given. skipping."
```
maralorn, line 222: The dropped mzero seems like a semantic change to me. Is that intentional?

maralorn (changes requested):
This is great actually! We should ship it.

I am just confused about one spot. Let’s clarify that.

maralorn (approved):

PR #649, opened 2025-02-27
Fix `cannot replace name binding ... Identifier "null"`

Mr-Andersen:
This is the error I encountered:

```
cabal2nix: post-process: cannot replace name binding Bind (Identifier "gstreamer") (Path [Identifier "null"]) by Bind (Identifier "gstreamer") (Path [Identifier "pkgs",Identifier "gst_all_1",Identifier "gstrea>
CallStack (from HasCallStack):
  error, called at src/Distribution/Nixpkgs/Haskell/FromCabal/PostProcess.hs:211:27 in cabal2nix-2.19.1-2Xhk8y7AoxGDHZPBS0CWBa:Distribution.Nixpkgs.Haskell.FromCabal.PostProcess
```

The second commit fixes it. I in fact don't understand this code and why this commit fixes it :) I use this version of cabal2nix for a personal project with ton of (indirect) dependencies, so it seems like it didn't break anything.

The first commit adds `HasCallStack` that I used in hope to understand what's going on here. I've left it in because I felt that callstack is really useful for exception.

### DIFF
--- a/cabal2nix/src/Cabal2nix.hs
+++ b/cabal2nix/src/Cabal2nix.hs
@@ -32,6 +32,7 @@ import Distribution.PackageDescription ( mkFlagName, mkFlagAssignment, FlagAssig
 import Distribution.Parsec as P
 import Distribution.Simple.Utils ( lowercase )
 import Distribution.System
+import GHC.Stack (HasCallStack)
 import Language.Nix
 import Options.Applicative
 import Paths_cabal2nix ( version )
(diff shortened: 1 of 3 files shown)


Mr-Andersen:
>How are you triggering this problem exactly

I've included `gi-gst` in .cabal file. This is the project: https://codeberg.org/aka_dude/web-tv. If you comment out line 44, you'll get this post-process error.

sternenseemann:
Thanks, that helps. I'll propose a fix soon.

sternenseemann:
https://github.com/NixOS/cabal2nix/pull/650 should fix the problem you were experiencing.

Mr-Andersen:
Yep, it fixes my issue, thank you.

PR #655, opened 2025-04-21
.github: run on ubuntu-24.04

wolfgangwalther:
The ubuntu-20.04 runners have been deprecated and disabled.

Side note: A bit annoying, that `haskell-ci` is so outdated on hackage and thus in nixpkgs. The dev shell for cabal2nix provides it, but it's way too old, so I had to build it from source. Tried updating it in nixpkgs to an unstable version, but didn't succeed, yet.

### DIFF
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -8,9 +8,9 @@
 #
 # For more information, see https://github.com/haskell-CI/haskell-ci
 #
-# version: 0.19.20241121
+# version: 0.19.20250330
 #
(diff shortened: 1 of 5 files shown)


sternenseemann:
I used to maintain a haskell-ci-unstable attribute, then they released frequently for a bit, but not anymore (for quite a while). Maybe I should bring it back. Unfortunately, the whole package is a massive pain, but I'm not sure about alternatives for a nixpkgs independent CI for Haskell (which is quite useful for cabal2nix).

sternenseemann:
What do you think? Should we maintain an unstable haskell-ci again?

PR #656, opened 2025-04-23
cabal2nix/test: support Cabal >= 3.14

sternenseemann:
We don't pass a CWD to readGenericPackageDescription since the whole test suite uses the working directory of the process to find the test data anyways (which we could improve in principle…).


sternenseemann:
> Should we add GHC 9.12 for CI at the same time to confirm it builds there as well?

Done, I've made the haskell-ci changes in this PR as well now.

sternenseemann:
Done, kept HLS for now since I think editors will struggle to find it otherwise.

PR #657, opened 2025-04-30
Add name mapping for python3-embed to python3

mrcjkb:
See https://github.com/NixOS/nixpkgs/pull/403018#pullrequestreview-2806649524


maralorn:
Nice.

Next steps: bump cabal2nix-unstable in nixpkgs (there is a script in maintainers/scripts/haskell for that)

PR #660, opened 2025-06-09
Fix warnings

alexfmpe:
Trying to use `ghcid -c 'cabal repl all --enable-multi-repl'` gives me a wall of warnings that really gets in the way.

Should we make .cabal and/or CI stricter so they don't accumulate again?


maralorn (approved):
Thank you!

PR #663, opened 2025-07-24
Add name mapping for libxxhash to xxHash

mrcjkb:
`haskellPackages.xxhash-ffi` (if the `pkg-config` flag is enabled) [depends on `libxxhash`](https://github.com/haskell-haskey/xxhash-ffi/blob/8cb38053efef67eb7c2a0d6d0d22eefbd7b1cf65/xxhash-ffi.cabal#L55), which is packaged as `xxHash` in nixpkgs.

Note: The build is actually broken because it can't find `pkg-config`:

```
Error: Setup: The program 'pkg-config' version >=0.9.0 is required but it
could not be found
```

So this is probably not enough to unbreak the package, but I think it will be necessary.


cabal2nix/src/Distribution/Nixpkgs/Haskell/FromCabal/Name.hs
```diff
@@ -140,6 +140,7 @@ libNixName "libxml-2.0"                         = return "libxml2"
 libNixName "libzip"                             = return "libzip"
 libNixName "libzmq"                             = return "zeromq"
 libNixName "liquid"                             = return "liquid-dsp"
+libNixName "libxxhash"                          = return "xxHash"
```
sternenseemann, line 143: Please make sure this list remains alphabetically ordered.

mrcjkb:
Thanks for the heads-up :pray: 
I guess it will be marked as unbroken with the next hackage-packages update?
If so, I'm in no rush. Otherwise, how do I remove the broken flag without updating all the packages? I'm assuming there's a maintainers script for that?

sternenseemann:
I expect someone will remove the stale broken flags semi-automatically this rotation, so you can probably just wait. Otherwise you'd remove the offending lines from `configuration-hackage2nix/broken.yaml` and run `maintainers/scripts/haskell/regenerate-hackage-packages.sh`.

PR #664, opened 2025-08-03
FromCabal.Name: more libNixName mappings

sternenseemann:


roberth:
I'm co-responsible for the `defaultPkgConfigPackages` mapping in Nixpkgs.
It'd be a good fallback, but I'm not sure that it's a complete solution.
Specifically some packages contain multiple pkg-config modules, and it'd be preferable to consume those through a single package function argument attribute, so that a single override changes the whole group of modules.
Otherwise, we risk that an incomplete override of some of the pkg-config modules causes for example an ABI compatibility problem.

roberth:
Something better could be created, but the interesting stuff won't happen within the broad `callPackage` area.
Probably the mapping in Nixpkgs needs to be more first-class instead of the already resolved attribute selections of `defaultPkgConfigModules`.
I'll "move" this to https://github.com/NixOS/nixpkgs/issues/273815
Thanks for the merge!

PR #665, opened 2025-08-27
cabal2nix: use inNixShell auto-call argument over environment var

sternenseemann:
The set environment variable leaks relatively easily leading to unexpected behavior when, for example, running nix-build from a shell or direnv environment. Nix 2.4 has provided an alternative mechanism for this which doesn't exhibit this flaw.

https://github.com/NixOS/nix/pull/3168
https://github.com/NixOS/nixpkgs/issues/436814
https://github.com/NixOS/nixpkgs/issues/9438


sternenseemann:
Unless people have `assert builtins.length (builtins.attrNames (builtins.functionArgs (import ./shell.nix))) == 3;` in their expressions, we're probably fine.

maralorn:
Really? I thought this autocalling only works when the function is the attribute passed to nix-shell. When however someone has a wrapper around their default.nix file, e.g. to pass in a pinned nixpkgs they would need to modify that wrapper to pass through that argument. don’t they?

PR #668, opened 2025-09-08
FromCabal: remove LLVM evil

emilazy:


doc/frequently-asked-questions.rst
```diff
@@ -488,7 +488,7 @@ The solution is to add it as another package to the environment:
 
 .. code:: shell
 
-   $ nix-shell -p '(haskellPackages.ghcWithPackages.override { useLLVM = true; }) (ps: with ps; [mtl])' 'haskellPackages.llvmPackages.llvm'
+   $ nix-shell -p '(haskellPackages.ghcWithPackages.override { useLLVM = true; }) (ps: with ps; [mtl])' 'haskellPackages.ghc.llvmPackages.llvm'
```
sternenseemann, line 491: This is fair since it is about maknig LLVM tools matching the LLVM backend available.

cabal2nix/src/Distribution/Nixpkgs/Haskell/FromCabal/PostProcess.hs
```diff
@@ -112,7 +112,7 @@ hooks =
   , ("git-annex >= 6.20170925 && < 6.20171214", set doCheck False)      -- some versions of git-annex require their test suite to be run inside of a git checkout
   , ("github-backup", set (executableDepends . tool . contains (pkg "git")) True)
   , ("GLFW", over (libraryDepends . system) (Set.union (Set.fromList [bind "pkgs.xorg.libXext", bind "pkgs.xorg.libXfixes"])))
-  , ("GlomeVec", set (libraryDepends . pkgconfig . contains (bind "self.llvmPackages.llvm")) True)
+  , ("GlomeVec", set (libraryDepends . pkgconfig . contains (bind "pkgs.llvmPackages.llvm")) True)
```
sternenseemann, line 115: This package passes `-fllvm`, so this is actually correct and necessary to bring the LLVM tools into scope.

sternenseemann (changes requested):

emilazy:
Sounds good.

PR #671, opened 2025-09-16
Nixpkgs.Haskell.FromCabal.Name: drop mapping for IL

emilazy:
As part of dropping it from Nixpkgs.


emilazy:
Well, I already applied the output in the Nixpkgs PR, so as long as this is merged before the next `cabal2nix` run it should be okay, I think.

PR #677, opened 2025-10-18
remove cases where cabal2nix produces a stringy license

jopejoe1:
I'm working on removing all stringy licenses in nixpkgs, and currently the only ones left are the ones in haskellPackages ~and node2nix packages~. This change fixes the ones for Haskell packages.

### DIFF
--- a/cabal2nix/src/Distribution/Nixpkgs/Haskell/FromCabal/License.hs
+++ b/cabal2nix/src/Distribution/Nixpkgs/Haskell/FromCabal/License.hs
@@ -41,7 +41,7 @@ fromCabalLicense AllRightsReserved                         = Known "lib.licenses
 fromCabalLicense (Apache Nothing)                          = Known "lib.licenses.asl20"
 fromCabalLicense (Apache (Just (versionNumbers -> [2,0]))) = Known "lib.licenses.asl20"
 fromCabalLicense ISC                                       = Known "lib.licenses.isc"
-fromCabalLicense OtherLicense                              = Unknown Nothing
+fromCabalLicense OtherLicense                              = Known "[ ]"
 fromCabalLicense (UnknownLicense "CC0-1.0")                = Known "lib.licenses.cc0"
 fromCabalLicense (UnknownLicense "BSD3ClauseORApache20")   = Known "lib.licenses.bsd3"
 fromCabalLicense l                                         = error $ "Distribution.Nixpkgs.Haskell.FromCabal.License.fromCabalLicense: unknown license"
@@ -63,7 +63,7 @@ fromSPDXExpression (SPDX.EAnd expres1 expres2) = Known ("lib.licenses.AND [ (" +
 fromSPDXExpression (SPDX.EOr expres1 expres2) = Known ("lib.licenses.AND [ (" ++ NPP.prettyShow (fromSPDXExpression expres1) ++ ") (" ++ NPP.prettyShow (fromSPDXExpression expres2) ++ ") ]")
 
 fromSPDXLicense :: SPDX.License -> Distribution.Nixpkgs.License.License
-fromSPDXLicense SPDX.NONE = Unknown Nothing
+fromSPDXLicense SPDX.NONE = Known "[ ]"
 fromSPDXLicense (SPDX.License expr) = fromSPDXExpression expr
 
 -- "isFreeLicense" is used to determine whether we generate a "hydraPlatforms =
--- a/cabal2nix/src/Distribution/Nixpkgs/Haskell/FromCabal/License.hs
+++ b/cabal2nix/src/Distribution/Nixpkgs/Haskell/FromCabal/License.hs
@@ -43,7 +43,7 @@ fromCabalLicense (Apache (Just (versionNumbers -> [2,0]))) = Known "lib.licenses
 fromCabalLicense ISC                                       = Known "lib.licenses.isc"
 fromCabalLicense OtherLicense                              = Known "[ ]"
 fromCabalLicense (UnknownLicense "CC0-1.0")                = Known "lib.licenses.cc0"
-fromCabalLicense (UnknownLicense "BSD3ClauseORApache20")   = Known "lib.licenses.bsd3"
+fromCabalLicense (UnknownLicense "BSD3ClauseORApache20")   = Known "lib.licenses.OR [ lib.licenses.bsd3 lib.licenses.asl20 ]"
 fromCabalLicense l                                         = error $ "Distribution.Nixpkgs.Haskell.FromCabal.License.fromCabalLicense: unknown license"
                                                                   ++ show l ++"\nChoose one of: " ++ intercalate ", " (map display knownLicenses)
 
--- a/cabal2nix/src/Distribution/Nixpkgs/Haskell/FromCabal/License.hs
+++ b/cabal2nix/src/Distribution/Nixpkgs/Haskell/FromCabal/License.hs
@@ -9,7 +9,8 @@ module Distribution.Nixpkgs.Haskell.FromCabal.License
 import Data.List (intercalate)
 import Distribution.License ( License(..), knownLicenses )
 import Distribution.Nixpkgs.License
-import Distribution.Pretty (prettyShow)
+import Distribution.Pretty as DP
+import Language.Nix.PrettyPrinting as NPP
 import qualified Distribution.SPDX as SPDX
 import Distribution.Text (display)
 import Distribution.Version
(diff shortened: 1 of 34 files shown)


cabal2nix/src/Distribution/Nixpkgs/Haskell/FromCabal/License.hs
```diff
@@ -40,128 +41,33 @@ fromCabalLicense AllRightsReserved                         = Known "lib.licenses
 fromCabalLicense (Apache Nothing)                          = Known "lib.licenses.asl20"
 fromCabalLicense (Apache (Just (versionNumbers -> [2,0]))) = Known "lib.licenses.asl20"
 fromCabalLicense ISC                                       = Known "lib.licenses.isc"
-fromCabalLicense OtherLicense                              = Unknown Nothing
+fromCabalLicense OtherLicense                              = Known "lib.licenses.unfree"
```
sternenseemann: This is incorrect. We assume all licenses are not unfree since Hackage requires open source licenses.

cabal2nix/src/Distribution/Nixpkgs/Haskell/FromCabal/License.hs
```diff
@@ -9,24 +9,25 @@ module Distribution.Nixpkgs.Haskell.FromCabal.License
...
+import Language.Nix.PrettyPrinting (prettyShow)
 import qualified Distribution.SPDX as SPDX
 import Distribution.Text (display)
 import Distribution.Version
 
 -- TODO: Programmatically strip trailing zeros from license version numbers.
 
 fromCabalLicense :: Distribution.License.License -> Distribution.Nixpkgs.License.License
-fromCabalLicense (GPL Nothing)                             = Unknown (Just "GPL")
+fromCabalLicense (GPL Nothing)                             = Known "lib.licenses.gpl2Plus"
```
sternenseemann: These cases are intentional as there is not enough information to conclude what variant of the license in question is used.

cabal2nix/src/Distribution/Nixpkgs/Haskell/FromCabal/License.hs
```diff
@@ -40,128 +41,33 @@ fromCabalLicense AllRightsReserved                         = Known "lib.licenses
...
-              Unknown (Just $ prettyShow expr)
-        _ ->
-          -- Not handed: the '+' suffix and user-defined licences references.
-          -- Use the SPDX expression as a free-form license string.
-          Unknown (Just $ prettyShow expr)
-    _ ->
-      -- Not handled: compound expressions, not expressible in Nixpkgs.
-      -- Use the SPDX expression as a free-form license string.
-      Unknown (Just $ prettyShow expr)
+fromSPDXLicense SPDX.NONE = Known "lib.licenses.unfree"
```
sternenseemann: Again, we simply conclude this has no specified license, but is open source.

cabal2nix/src/Distribution/Nixpkgs/Haskell/FromCabal/License.hs
```diff
@@ -40,128 +41,33 @@ fromCabalLicense AllRightsReserved                         = Known "lib.licenses
...
+    -- Not handed: the '+' suffix and user-defined licences references.
+    -- Use the SPDX expression as a free-form license string.
+    Unknown (Just $ DPretty.prettyShow simpl)
+fromSPDXLicenseExpression (SPDX.ELicense simpl (Just excep)) =
+  case simpl of
+  SPDX.ELicenseId lid -> Known ("lib.licensesSpdx.\"" ++ DPretty.prettyShow lid ++ "\" lib.licensesSpdx.\"" ++  DPretty.prettyShow excep ++ "\"")
+  _ ->
+    -- Not handed: the '+' suffix and user-defined licences references.
+    -- Use the SPDX expression as a free-form license string.
+    Unknown (Just $ DPretty.prettyShow simpl ++ " WITH " ++ DPretty.prettyShow excep)
```
sternenseemann: Why are we reconstructing the entire expression here? We can just `prettyShow` the entire spdx expr.

sternenseemann (changes requested):
Please also add a changelog entry, ideally noting when nixpkgs added support for `lib.licensesSpdx`.

cabal2nix/src/Distribution/Nixpkgs/Haskell/FromCabal/License.hs
jopejoe1: From what I can see at this point in the function, I do not have access to the full spdx expr, but maybe I'm missing something. This was my first time touching Haskell, still quite inexperienced with the language.

cabal2nix/src/Distribution/Nixpkgs/Haskell/FromCabal/License.hs
jopejoe1: Same as above, will switch this to "lib.licenses.free"

cabal2nix/src/Distribution/Nixpkgs/Haskell/FromCabal/License.hs
jopejoe1: I do realize they were intentional, but they can't stay in their current form. I'm currently working towards disallowing stringy licenses in nixpkgs, and those are pretty much the only ones left. Other options would be to make it even less informational with `lib.licenses.free` or to add a new custom license (not really a fan of this).

cabal2nix/src/Distribution/Nixpkgs/Haskell/FromCabal/License.hs
```diff
@@ -40,128 +41,33 @@ fromCabalLicense AllRightsReserved                         = Known "lib.licenses
...
+  SPDX.ELicenseId lid -> Known ("lib.licensesSpdx.\"" ++ DPretty.prettyShow lid ++ "\"")
+  _ ->
+    -- Not handed: the '+' suffix and user-defined licences references.
+    -- Use the SPDX expression as a free-form license string.
+    Unknown (Just $ DPretty.prettyShow simpl)
+fromSPDXLicenseExpression (SPDX.ELicense simpl (Just excep)) =
+  case simpl of
+  SPDX.ELicenseId lid -> Known ("lib.licensesSpdx.\"" ++ DPretty.prettyShow lid ++ "\" lib.licensesSpdx.\"" ++  DPretty.prettyShow excep ++ "\"")
+  _ ->
+    -- Not handed: the '+' suffix and user-defined licences references.
```
wolfgangwalther:
```suggestion
    -- Not handled: the '+' suffix and user-defined licences references.
    -- Use the SPDX expression as a free-form license string.
    Unknown (Just $ DPretty.prettyShow simpl)
fromSPDXLicenseExpression (SPDX.ELicense simpl (Just excep)) =
  case simpl of
  SPDX.ELicenseId lid -> Known ("lib.licensesSpdx.\"" ++ DPretty.prettyShow lid ++ "\" lib.licensesSpdx.\"" ++  DPretty.prettyShow excep ++ "\"")
  _ ->
    -- Not handled: the '+' suffix and user-defined licences references.
```

cabal2nix/test/golden-test-cases/highlighting-kate.nix.golden
```diff
@@ -18,5 +18,5 @@ mkDerivation {
   ];
   homepage = "http://github.com/jgm/highlighting-kate";
   description = "Syntax highlighting";
-  license = "GPL";
+  license = lib.licenses.gpl2Plus;
```
wolfgangwalther: Upstream still uses `GPL`, and has `gpl2Only`, I think: https://github.com/jgm/highlighting-kate/blob/master/LICENSE

cabal2nix/test/golden-test-cases/HTF.nix.golden
```diff
@@ -27,6 +27,6 @@ mkDerivation {
   ];
   homepage = "https://github.com/skogsbaer/HTF/";
   description = "The Haskell Test Framework";
-  license = "LGPL";
+  license = lib.licenses.lgpl2Plus;
```
wolfgangwalther: Upstream has now `LGPL-2.1`, which renders as `LGPL-2.1-only` on hackage.

cabal2nix/src/Distribution/Nixpkgs/Haskell/FromCabal/License.hs
sternenseemann: There is also a typo in the commit message.

jopejoe1:
Sorry for not being really active in here was working on an improvement for licenses in nixpkgs for making this nicer to implement here. see https://github.com/NixOS/nixpkgs/pull/468378

jopejoe1:
Did a larger refactor based on the changes done in https://github.com/NixOS/nixpkgs/pull/468378 and the feedback that was given here.

cabal2nix/src/Distribution/Nixpkgs/Haskell/FromCabal/License.hs
```diff
@@ -40,128 +41,33 @@ fromCabalLicense AllRightsReserved                         = Known "lib.licenses
...
+fromSPDXLicenseExpression  (SPDX.ELicense simpl Nothing) =
+  case simpl of
+  SPDX.ELicenseId lid -> Known ("lib.licensesSpdx.\"" ++ DPretty.prettyShow lid ++ "\"")
+  _ ->
+    -- Not handed: the '+' suffix and user-defined licences references.
+    -- Use the SPDX expression as a free-form license string.
+    Unknown (Just $ DPretty.prettyShow simpl)
+fromSPDXLicenseExpression (SPDX.ELicense simpl (Just excep)) =
+  case simpl of
+  SPDX.ELicenseId lid -> Known ("lib.licensesSpdx.\"" ++ DPretty.prettyShow lid ++ "\" lib.licensesSpdx.\"" ++  DPretty.prettyShow excep ++ "\"")
```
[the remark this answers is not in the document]
jopejoe1: This now always returns a valid SPDX expression; this was not possible before, as we are not allowed to have nested lists in `meta.license`

jopejoe1:
Now that compound licenses are supported in all supported nixpkgs versions, could this be moved forward?

sternenseemann:
Yes, I'd be interested in that!

Though, I'm still of the position that we should not fall back to something nondescript like `licenses.free` if we don't recognize the license. That can be solved manually in Nixpkgs or upstream (ideally). Probably a lot of packages affected by this would also wind up [excluded](https://github.com/NixOS/cabal2nix/pull/667) eventually.

jopejoe1:
Should work like this now.

Issue #550, opened 2022-05-02
cabal2nix tries to fetch hackage revisions even if already present in the source


sternenseemann:
Consider the [servant-cassava.cabal file](https://github.com/haskell-servant/servant-cassava/blob/66617547851d38d48f5f1d1b786db1286bdafa9d/servant-cassava.cabal#L4): It has a Hackage `x-revision` field checked in the repository which, interestingly, was never applied on Hackage.

`cabal2nix`, however, recognizes this and sets the `revision` and even `editedCabalFileSha256` (I have no clue yet where this hash comes from). Arguably, when `cabal2nix`ing a revision of a git repo where the modifications of the revision are checked in, it doesn't make sense to fetch the revision from Hackage (it may even fail as in this case). Fixing this will require careful checking if there are any cases where this would be desireable and, most importantly, how this works exactly in `all-cabal-hashes`, so we don't regress `hackage2nix`.

Ericson2314:
I don't think the old behavior is ever desirable. When cabal2nix is given source code it shouldn't be doing IO to get other source code, full stop. That just doesn't make sense at all. In general, don't mix IO with business logic like that too.

sternenseemann:
There is some nuance to it, since there is a difference between being given a cabal file and being given source code. This difference is already present in the fetcher code, so I think we just need to mirror it for the revision logic.

sternenseemann:
I think the current behavior is actually correct. As discussed on https://github.com/NixOS/cabal2nix/pull/715, when processing a revised cabal file from downloaded from Hackage directly, cabal2nix generates the correct expression.

I think the answer is simply that it does not make sense for source repositories to contain cabal files with `x-revision`. (It will still work if they match an actual Hackage revision *exactly*, though.)

Possibly, the fetcher logic should separate the way to fetch the source and the source of the cabal file further.

PR #715, opened 2026-04-02
Add `--cabal-file-{sha256,revision}` arguments

9999years:
This lets `cabal2nix` generate derivations with `editedCabalFile` and `revision` fields. This functionality already exists in `hackage2nix` but wasn't exposed in `cabal2nix`.

### DIFF
--- a/cabal2nix/src/Cabal2nix.hs
+++ b/cabal2nix/src/Cabal2nix.hs
@@ -54,6 +54,8 @@ data Options = Options
   , optDoBenchmark :: Bool
   , optRevision :: Maybe String
   , optHyperlinkSource :: Bool
+  , optCabalFileSha256 :: Maybe String
+  , optCabalRevision :: Maybe Int
   , optEnableLibraryProfiling :: Bool
   , optEnableExecutableProfiling :: Bool
   , optEnableProfiling :: Maybe Bool
@@ -99,6 +101,10 @@ options = do
     <- optional (strOption $ long "revision" <> help "revision to use when fetching from VCS")
   optHyperlinkSource
     <- flag True False (long "no-hyperlink-source" <> help "don't generate pretty-printed source code for the documentation")
+  optCabalFileSha256
+    <- optional (strOption $ long "cabal-file-sha256" <> metavar "HASH" <> help "sha256 hash of the revised .cabal file (sets editedCabalFile in output)")
+  optCabalRevision
+    <- optional (option auto $ long "cabal-file-revision" <> metavar "N" <> help "revision number of the .cabal file (sets revision in output)")
   optEnableLibraryProfiling
     <- switch (long "enable-library-profiling" <> help "enable library profiling in the generated build")
   optEnableExecutableProfiling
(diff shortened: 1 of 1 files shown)


sternenseemann:
This functionality is exposed in `cabal2nix`, but only in certain situations:

- When using the cabal-install database, it will correctly be generated based on the current DB state, e.g. `cabal2nix cabal://socket-0.8.3.0`.
- If the cabal file has a `x-revision` field (like Hackage would have), cabal2nix will also generate the arguments, though I would need to check whether the `editedCabalFileSha256` is actually correct in that case. See also https://github.com/NixOS/cabal2nix/issues/550. I'm not sure what the best sort of behavior is in that situation, realistically, I think, cabal files not downloaded from Hackage shouldn't have `x-revision` set.

Maybe you could elaborate why these extra flags are interesting to you? I think revision are a little bit tricky. Ideally, we would just have cabal2nix do the right thing when necessary and otherwise hide this detail from the user as much as possible. (Arguably the mechanism is a layer violation to begin with which is probably why dealing with it is so annoying…)

9999years:
> If the cabal file has a `x-revision` field (like Hackage would have), cabal2nix will also generate the arguments

Ah, that might be it; I think I tested this with revision 0 of a `.cabal` file.

For some more context, this is related to the "`vendor-cabal2nix`" project which we intend to use to pin various Haskell dependencies in an ergonomic fashion without incurring IFD costs.

9999years:
Yes, that works! Thanks for the reminder. I can close this, apologies for the ping.

```bash
curl -O https://hackage.haskell.org/package/aeson-diff-1.1.0.13/revision/1.cabal
cabal2nix --sha256 8b0d0b9dd58675124d85c9285dce7028c4c9246b46ba5eaad2610269011ea169 1.cabal
```

```nix
{ mkDerivation, aeson, base, bytestring, Cabal, cabal-doctest
, directory, doctest, edit-distance-vector, filepath, Glob, lib
, optparse-applicative, QuickCheck, scientific, text, vector, yaml
}:
mkDerivation {
  pname = "aeson-diff";
  version = "1.1.0.13";
  sha256 = "8b0d0b9dd58675124d85c9285dce7028c4c9246b46ba5eaad2610269011ea169";
  revision = "1";
  editedCabalFile = "1028adallw7bm72948lj322bb5a99gfs0qc1j0pnm8hryp6n7ma5";
  isLibrary = true;
  isExecutable = true;
  setupHaskellDepends = [ base Cabal cabal-doctest ];
  libraryHaskellDepends = [
    aeson base bytestring edit-distance-vector scientific text vector
  ];
  executableHaskellDepends = [
    aeson base bytestring optparse-applicative yaml
  ];
  testHaskellDepends = [
    aeson base bytestring directory doctest filepath Glob QuickCheck
    vector
  ];
  homepage = "https://github.com/ysangkok/aeson-diff";
  description = "Extract and apply patches to JSON documents";
  license = lib.licenses.bsd3;
}
```

sternenseemann:
Is the `editedCabalFile` sha256 actually correct, did you check?

9999years:
Looks good:

```
$ curl -O https://hackage.haskell.org/package/aeson-diff-1.1.0.13/revision/1.cabal
$ sha256sum 1.cabal
45d563cdf519a26a2f908161a0dd4b4995b58418922292c4a9eb704a55534880  1.cabal
$ nix-hash --to-base16 --type sha256 1028adallw7bm72948lj322bb5a99gfs0qc1j0pnm8hryp6n7ma5
45d563cdf519a26a2f908161a0dd4b4995b58418922292c4a9eb704a55534880
```

sternenseemann:
Thanks for checking, great!

PR #720, opened 2026-08-16
Check `nix-instantiate` exit code in

bglgwyng:
`readNixpkgSet` runs `nix-instantiate --strict ...` to obtain the package map for nixpkgs.

Previously, if `nix-instantiate` failed(for example, due to the absense of Nix CLI tools), the error was silently ignored, and the function returned an empty set.

This change checks the process exit code and reports failures properly.


sternenseemann:
This looks good, thanks! I'll add a changelog entry and probably merge this into the next breaking release of `distribution-nixpkgs`, just to be safe.

